PR #10, opened 2018-07-29
Split out RGB conversion and encourage inlining

hcs64:
I noticed that drawing images in Plato seemed a bit slow, it looks like this is partly due to the compiler not inlining pixel writing, resulting in a few function calls in every pixel! The changes I've attached here seem to satisfy the compiler.


canselcik:
Merged. I appreciate the PR. Feel free to open more PRs or create issues as you come across them.

Issue #5, opened 2018-03-06
Suggestion: use "config.mk" model for easily customizing build flags


PaulBatchelor:
A configuration file for Make would be a nice thing to have. I thought I would run it by you before making a PR.

For me, this would be necessary because my cross-compiler toolchains are in weird, non-standard locations. 

The classic example of this is with [dwm](https://github.com/sr/dwm). I would probably implement something similar to my own projects [Soundpipe](https://github.com/paulbatchelor/soundpipe) and [Sporth](https://github.com/paulbatchelor/sporth). 

What do you think?

canselcik:
Yeah I wouldn't be against it. You can definitely focus on improving the config and build steps of the C implementation. There also might be some value in making it download and extract the Poky Toolchain before a build, or at least check if `arm-linux-gnueabihf-gcc` exists and can be used to build executables.

PaulBatchelor:
>  download and extract Poky Toolchain before a build

It would be cleaner to put this in a shell script that the user explicitly calls rather than try to have Make do  that.

> or at least check if arm-linux-gnueabihf-gcc exists

The `?=` operator should do the job, as it will set `CC` to that value by default unless overwritten. A config file could override those defaults too.

PatrickGretzki:
Its seems, that the "nightly" channel is needed for compiling the rust example ... otherwise "mock_derive" won't compile

canselcik:
@PatrickGretzki Thanks for pointing that out. All the `mock`ing code was introduced with https://github.com/canselcik/libremarkable/commit/517a0445ecdce63df4eacd6d60904286ee7d559c, you can do `git revert 517a044` as a temporary workaround (or use the nightly channel) as we aren't even leveraging these mocking capabilities yet.

Let me know if you have further comments/suggestions, especially since `PatrickGretzki/paperOS` looks like we have quite an overlap there. :)

PatrickGretzki:
I just started using rust, so I will have many questions in the next days :)
Otherwise regarding "PaperOS" - in the moment I try to collect ideas and features for a smart replacement of "xochitl". As soon as the concept is ready and I tested some design issues, we should work together to make it reality.

PatrickGretzki:
Hey. its me again- The cross-compiler is running and I can compile simple programs for the remarkable. But  building your example/lib is still not working. The package "lua52-sys" throws a panic error at compiling 
`Compiling lua52-sys v0.1.1`
[...]

PR #11, opened 2018-09-08
Docker toolchain

SafariMonkey:
This PR adds a docker image specification that can be used to build rust binaries for the reMarkable. It works on Linux (untested on OS X) and only requires Docker and make as dependencies.

Before this PR, a user compiling on ubuntu 16.04 would need to perform all the steps listed [here](https://github.com/canselcik/libremarkable/issues/5#issuecomment-377592830) except for starting a VM. On 18.04, I was unable to cross compile successfully at all due to glibc versioning, despite much effort. However, with this PR merged, any linux user with docker, make and reMarkable ssh set up can do:
```
git clone https://github.com/canselcik/libremarkable
cd libremarkable
make run-docker
```
and see the demo running on their device. They can also copy the make target for compiling their own programs.

To make it easier to build in docker and deploy directly, the `run` target is now an alias for `examples` and `deploy-demo`. There is a new target, `run-docker`, which builds the binary in docker instead of locally before deploying. This pattern can be applied to other deployment targets if desired.

It also makes two changes that were needed for a successful build on latest nightly: pinning `atomic` to 0.4.0, and removing the `const_size_of` feature gate (as it appears to have been removed)

A couple technical notes:

- It's slower, but not that much.
    - The first build of the docker image and binary are expected to take some time. However, all three slow steps are cached:
        - The docker build is cached by docker, and should rarely change
        - The rust registry is mounted in a volume, and thus persisted across builds
        - The `target` folder is on the host, so the compiler cache is persisted across builds.
    - Some sample build times on my machine:
        - First build, no caches: `7m17.09s`
        - Build only changing `demo.rs`: `0m8.28s`
        - Build with no changes: `0m3.24s`
- Docker layer cache may cause the rust toolchain to go out of date. A couple potential workarounds:
    - periodically build the image with cache disabled to force a redownload
    - invalidate the cache by adding a file containing the date
- Mounting and using the directory without affecting permissions requires setting the UID and GID of the user inside the container to the owner of the directory. For now it is set to the UID and GID of the user running the make command, which should work in most cases. This also means the image is specific to the UID-GID combination, and can't easily be distributed. There are methods that solve that problem, like [what redis does](https://stackoverflow.com/a/29799703).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ hlua = "0.4.1"
 aabb-quadtree = "0.1.0"
 zstd = "0.4.14"
 stopwatch = "0.0.7"
-atomic = {version = "0.4", features = ["nightly"]}
+atomic = {version = "= 0.4.0", features = ["nightly"]}
 
 [features]
 enable-runtime-benchmarking = []
--- a/docker-toolchain/.cargo/config
+++ b/docker-toolchain/.cargo/config
@@ -1,0 +1,2 @@
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,25 @@ examples:
 bench:
 	cargo build --examples --release --target=armv7-unknown-linux-gnueabihf --features "enable-runtime-benchmarking"
 
+.PHONY: docker-env
+docker-env:
+	cd docker-toolchain && docker build \
+		--build-arg UNAME=builder \
+		--build-arg UID=$(shell id -u) \
+		--build-arg GID=$(shell id -g) \
+		--tag rust-build-remarkable:latest .
+
+examples-docker: docker-env
+	docker volume create cargo-registry
+	docker run \
+		--rm \
+		--user builder \
+		-v $(shell pwd):/home/builder/libremarkable:rw \
+		-v cargo-registry:/home/builder/.cargo/registry \
+		-w /home/builder/libremarkable \
+		rust-build-remarkable:latest \
+		cargo build --examples --release --target=armv7-unknown-linux-gnueabihf
+
 library:
 	cargo build --release --target=armv7-unknown-linux-gnueabihf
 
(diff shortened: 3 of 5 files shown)


docker-toolchain/.cargo/config
```diff
@@ -0,0 +1,2 @@
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
```
canselcik: Not sure if the default `rustflags` imply this already or if this even makes a difference in terms of runtime performance however I have seen this added for binaries targeting `imx6slevk`:
```
rustflags = [
   "-C", "target-feature=+v7,+vfp3,+a9,+neon"
]
```

SafariMonkey: You're welcome to make changes as necessary, as long as you test them. I don't know if that's in the scope of this PR, though, unless those flags are already being used in the existing toolchain. Still, it sounds like something to look into.

Cargo.toml
```diff
@@ -29,7 +29,7 @@ hlua = "0.4.1"
 aabb-quadtree = "0.1.0"
 zstd = "0.4.14"
 stopwatch = "0.0.7"
-atomic = {version = "0.4", features = ["nightly"]}
+atomic = {version = "= 0.4.0", features = ["nightly"]}
```
canselcik: This should read:
```
atomic = {version = "0.4.0", features = ["nightly"]}
```
instead, right?

SafariMonkey: Unfortunately not. According to [this document](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), `"0.4.0"` would include version `0.4.1`, which is the version I am trying to exclude as it seems to me to be broken on latest nightly. (I originally specified `"0.4.0"` and had issues because it was using `0.4.1`.) Try `0.4.1` if you like (make sure it's updated in your lock file) and see if it works for you on latest nightly. It's possible that something is wrong in my end.

docker-toolchain/.cargo/config
SafariMonkey: If you're already using them, and it works, it would make sense to me to either include them in this PR or add them shortly after. If you'd prefer, I could also add them from my end.

Updating documentation to include them (where they are missing) may also help.

PR #14, opened 2018-10-02
Use cgmath for most coordinate representations

SafariMonkey:
This PR implements use of `cgmath` ([github](https://github.com/rustgd/cgmath), [docs](https://docs.rs/cgmath)) in most places in the library (including the API) where floats or field/argument pairs have previously been used. An overview of the changes:

- use `cgmath` for all API coordinates, except `mxcfb_rect` and hardware translation layers
- add some utility methods to `cgmath` and use them
- change some numerical types, in an attempt to improve consistency and ergonomics:
  - expose float arguments where it makes sense
  - allow negative offsets in position, as this allows things to be drawn with an off-screen position that overlaps onto the screen.
  - try to use the same number type width (e.g. u16 vs u32) in the same context unless there is an obvious reason not to
    - to be honest, I'm not sure what the reasoning is behind the existing type choices, but there's quite a variety. I'd like feedback on this; I'm happy to revert if necessary, but I think paring down the set of types in use (especially in the public API) further would be great if possible.
- fix a potential bug in a position comparison: the check `if rect.top != y as u32 && rect.left != x as u32` will fail if either coordinate is the same.
  - This unfortunately means that UI elements do a full redraw much more frequently, including the text for brush size. (The actual `min_y` value of the bounding box is lower than the `y` of the nominal position, but the `x` is the same for both.)
- use `.cast().unwrap()` for casting element types. It's not clear when this can fail, and there is an [outstanding issue](https://github.com/rust-num/num-traits/issues/9) for documenting the possible failure conditions. However, I haven't seen any errors in my testing. If any panics due to this _are_ encountered, in my opinion that cast should be fixed and the use of `.cast().unwrap()` should be re-evaluated.

There may still be inconsistencies that need resolving. Those are likely either an oversight on my part or a reluctance to change types without a clear understanding of the reasoning behind them.

This PR is not yet thoroughly tested, but I wanted to get it out there so any issues with approach could be identified early on. After that stage, I think a good way to test it practically would be to port an existing, larger application using `libremarkable` already and see how it goes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ aabb-quadtree = "0.1.0"
 zstd = "0.4.14"
 stopwatch = "0.0.7"
 atomic = {version = "= 0.4.0", features = ["nightly"]}
+cgmath = "0.16"
 
 [features]
 enable-runtime-benchmarking = []
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -8,6 +8,7 @@ extern crate env_logger;
 
 #[macro_use]
 extern crate libremarkable;
+use libremarkable::framebuffer::cgmath;
 use libremarkable::framebuffer::common::*;
 use libremarkable::framebuffer::refresh::PartialRefreshMode;
 use libremarkable::framebuffer::storage;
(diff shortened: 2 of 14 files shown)


canselcik:
Hi @SafariMonkey,
This looks great as usual. At a quick glance I don't see anything wrong with it at all. 

I am going to be taking a deeper look later tonight especially the following two parts:

> fix a potential bug in a position comparison: the check if rect.top != y as u32 && rect.left != x as u32 will fail if either coordinate is the same.
> 
> This unfortunately means that UI elements do a full redraw much more frequently, including the text for brush size. (The actual min_y value of the bounding box is lower than the y of the nominal position, but the x is the same for both.)

> use .cast().unwrap() for casting element types. It's not clear when this can fail, and there is an outstanding issue for documenting the possible failure conditions. However, I haven't seen any errors in my testing. If any panics due to this are encountered, in my opinion that cast should be fixed and the use of .cast().unwrap() should be re-evaluated.

Thanks.

canselcik:
I've had a chance to test these changes on my device and I see no problems at all with the code at all. Thanks again @SafariMonkey. Your contributions are very much appreciated and looking forward to the other changes you have in the pipeline. :)

PR #15, opened 2018-10-07
Standardise on u32 and i32 instead of usize/isize where possible.

SafariMonkey:
This is a fairly find and replace change of all instances of isize/usize with i32/u32, except where we deal with memory directly. I discussed this possibility with @canselcik recently so I thought I'd put out a diff for further discussion.

There is exactly one place where I have left isize in the external API:

```
    /// Reads the value at offset `ofst` from the mmapp'ed framebuffer region
    fn read_offset(&self, ofst: isize) -> u8;
```

As this deals directly with a memory offset, I thought it best to leave it as is.

I also included a `cargo fmt`, but I can remove that if you prefer.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -37,11 +37,11 @@ use std::time::Duration;
 
 #[derive(Copy, Clone, PartialEq)]
 enum DrawMode {
-    Draw(usize),
-    Erase(usize),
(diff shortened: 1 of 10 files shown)


canselcik:
Looks great and just like we discussed earlier. Merging.

PR #16, opened 2018-10-14
Graphics rendering improvements

SafariMonkey:
Summary:
- add polygon rasterisation
- base bezier drawing on rasterisation
- add bezier method with width interpolation
- parameterise bezier drawing by sample count
- refactor graphics functions into separate package with testability
- change circle filling to use plain squares comparison
- return floats for pen positions
- smooth out pen drawing by using sample midpoints
- add diamond touch mode for demonstration of vector drawing
- add -- and ++ buttons for quick brush sizing (+/-10)

Points of note:
- This does not fix the text drawing issue introduced in the previous PR. That will require a separate investigation.
- This is a breaking API change, but less so than the cgmath change.
- By making the `graphics` module non-public, functions can be freely moved into it as necessary without API changes. Functionality it provides that is not possible externally (e.g. stamping arbitrary draw functions along a line) can be exposed later if desired as an API change.
- Given that there are a number of changes in this PR, if there is any change that you wish to be split out I can certainly attempt to do so. Some are more intertwined than others.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -561,7 +561,7 @@ fn on_touch_handler(app: &mut appctx::ApplicationContext, input: multitouch::Mul
                 m @ TouchMode::Diamonds | m @ TouchMode::FillDiamonds => {
                     let position_int = position.cast().unwrap();
                     framebuffer.draw_polygon(
-                        vec![
+                        &vec![
                             position_int + cgmath::vec2(-10, 0),
                             position_int + cgmath::vec2(0, 20),
                             position_int + cgmath::vec2(10, 0),
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -533,14 +533,28 @@ fn on_touch_handler(app: &mut appctx::ApplicationContext, input: multitouch::Mul
             let rect = match G_TOUCH_MODE.load(Ordering::Relaxed) {
                 TouchMode::Bezier => {
                     let position_float = position.cast().unwrap();
-                    framebuffer.draw_bezier(
-                        position_float,
-                        position_float + cgmath::vec2(155.0, 14.0),
-                        position_float + cgmath::vec2(200.0, 200.0),
-                        2.0,
-                        1000,
-                        color::BLACK,
-                    )
+                    let points = vec![
+                        (cgmath::vec2(-40.0, 0.0), 2.5),
+                        (cgmath::vec2(40.0, -60.0), 5.5),
+                        (cgmath::vec2(0.0, 0.0), 3.5),
+                        (cgmath::vec2(-40.0, 60.0), 6.5),
+                        (cgmath::vec2(-10.0, 50.0), 5.0),
+                        (cgmath::vec2(10.0, 45.0), 4.5),
+                        (cgmath::vec2(30.0, 55.0), 3.5),
+                        (cgmath::vec2(50.0, 65.0), 3.0),
+                        (cgmath::vec2(70.0, 40.0), 0.0),
+                    ];
+                    let mut rect = mxcfb_rect::invalid();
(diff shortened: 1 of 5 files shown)


canselcik:
Another great patch @SafariMonkey. I am taking a look at this right now.

examples/demo.rs
```diff
@@ -502,18 +530,39 @@ fn on_touch_handler(app: &mut appctx::ApplicationContext, input: multitouch::Mul
...
-                    position_float + cgmath::vec2(155.0, 14.0),
-                    position_float + cgmath::vec2(200.0, 200.0),
-                    2.0,
-                    color::BLACK,
-                ),
+                TouchMode::Bezier => {
+                    let position_float = position.cast().unwrap();
+                    framebuffer.draw_bezier(
+                        position_float,
+                        position_float + cgmath::vec2(155.0, 14.0),
```
canselcik: These values were quite arbitrary. Feel free to draw a more interesting curve here.

src/framebuffer/draw.rs
```diff
@@ -161,18 +78,16 @@ impl<'a> framebuffer::FramebufferDraw for core::Framebuffer<'a> {
...
-            {
-                self.write_pixel(
-                    Point2 {
-                        x: x as i32,
-                        y: y as i32,
-                    },
-                    v,
-                );
+    fn fill_circle(&mut self, pos: cgmath::Point2<i32>, rad: u32, v: color) -> mxcfb_rect {
+        let rad_square = (rad * rad) as i32;
```
canselcik: Much better than the previous hacky `for current_rad in { 1..rad + 1 } { draw_circle(.., current_rad) }` approach. I like it.

src/framebuffer/mod.rs
```diff
@@ -57,13 +59,31 @@ pub trait FramebufferDraw {
...
         &mut self,
         startpt: cgmath::Point2<f32>,
         ctrlpt: cgmath::Point2<f32>,
         endpt: cgmath::Point2<f32>,
         width: f32,
+        samples: i32,
+        v: common::color,
+    ) -> common::mxcfb_rect;
+    /// Draws a bezier curve begining at `startpt`, with control point `ctrlpt`, ending at `endpt`
+    /// with a width at each point and color `color`
```
canselcik: After seeing the implementation in the other file for `draw_dynamic_bezier`, I was going to say let's make sure we document how it allows individual widths for each point but looks like you've already taken care of that. :+1:

src/framebuffer/draw.rs
SafariMonkey: Worth noting that this isn't quite the same. It can be solved by adding a floating point multiplication to make the rad*rad approximately halfway to the next radius, so the error is spread evenly rather than always being negative.

examples/demo.rs
canselcik: Yeah, a lot more interesting for sure.

[As seen here](https://i.imgur.com/AK2cU3O.jpg)

SafariMonkey: That looks right to me! Interesting was my goal. I actually drew a scribble and then modelled the curve on that, to get a fairly natural flow.

Now, is that the kind of interesting you were looking for?

canselcik:
This looks good to me. Merging.

SafariMonkey:
> The only thing that keeps slightly bothering me is the next full refreshes on updating text elements after that bugfix. Especially when that refresh coincides with the user drawing.

I agree, and I may spend some time working on that. I already had a go, and while a quick fix (doing a dry run and checking position changes based on that) is pretty easy, a larger fix (that handles both size and position changes optimally) will be more work.

Let me know if you have any suggestions or would prefer to figure that one out yourself.

PR #19, opened 2019-02-25
Use epoll 4.0.1 to avoid memory corruption

hcs64:
With current `rustc 1.34.0-nighty (097c04cf4 2019-02-24)`, I'm getting crashes in `demo` if I hit the "Disable Touch" button and then touch the screen. I tracked this down with valgrind to an overwrite from a syscall due to a struct being the wrong size, this has been fixed in epoll 4.0.1. Bumping the version doesn't seem to break anything.

With this fixed valgrind complains about uninitialized padding inside of `Event`, doesn't seem worth addressing.

valgrind also throws up [an issue with evdev](https://github.com/cmr/evdev/issues/14), but it seems harmless.


canselcik:
Thank you.

PR #25, opened 2019-10-06
Closes #24

### LINKED ISSUE #24: Battery location has changed in 1.8
In 1.8 the battery location has changed to `/sys/class/power_supply/bq27441-0`

merqurio:
The path to the battery information changed with the latest update (1.8).
This works (I just tested it), but this would break the build for libremarkables < 1.8, so I open the PR to discuss the best solution.

Possible solutions that came to my mind:
- a) Notice it in the README as a breaking change
- b) Check 1.7 path and if not available use the 1.8 path

Bests

### DIFF
--- a/src/battery.rs
+++ b/src/battery.rs
@@ -5,7 +5,7 @@ use std::io::Read;
 
 fn read_attribute(attr: &str) -> Result<String, String> {
     let mut data = String::new();
-    match File::open(format!("/sys/class/power_supply/bq27441/{0}", attr)) {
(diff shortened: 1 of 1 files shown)


Eeems:
I'd suggest checking the 1.8 path first, log a warning if it's not found, and then use the 1.7 path.

merqurio:
Maybe it makes more sense to just try to open it and deal with the `Result` ?
If 1.8 is fails -->  try 1.7

Eeems:
That would also work

PR #28, opened 2020-04-19
Truely static builds on the host with `musl`

fenollp:
To be able use musl I had to:
* rely on `cross` (I failed to set up a Docker toolchain: --target is not enough for cargo...)
* change to a signed `framebuffer::common::NativeWidthType` as Rust's musl seems to require it
* enable link-time optimization otherwise `mmap` would fail complaining about unaligned memory.

Side changes:
* fixed up warnings about unused Cargo.toml fields and simplifiable Rust syntax
* added tweakable `DEVICE_HOST` and `TARGET` variables to Makefile
* upgraded the dependencies that could be
* various whitespace noise


README.md
```diff
@@ -63,3 +63,15 @@ If you choose to skip the `Makefile` and call `cargo` yourself, make sure to inc
...
+Use
+```
+[target.armv7-unknown-linux-musleabihf]
+linker = "arm-linux-gnueabihf-gcc"
+```
+```
+TARGET=armv7-unknown-linux-musleabihf make demo deploy-demo
+```
+so far:
+> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrUnaligned', src/framebuffer/core.rs:67:23
```
fenollp, line 77: @canselcik (or anyone) any idea why mmap would fail with musl but otherwise work with glibc at https://github.com/canselcik/libremarkable/blob/a71f389e970df79ce8fba826fc658df54e4d6dbf/src/framebuffer/core.rs#L67 ?

sandsmark: tried testing with using musl's mmap directly? (i. e. from c/c++ instead of through the rust mmap library/bindings)

fenollp: Hi! Thanks for dropping by :)
Turns out rustc's pre-linking optimization was resolving symbols wrong: using LTO fixes it.

fenollp:
It works!
I've updated the PR's description. Hopefully this helps.

@sandsmark Are there any Rust programs at reMarkable AS?

canselcik:
Great work @fenollp, thank you for getting to the bottom of the crashes when linked against musl.

I actually had experimented with truly static builds in this repo before as well and had the same kind of crashes on mmap.

Interesting that enabling LTO fixes it. This looks great. I wonder what people will say in https://github.com/rust-embedded/cross/issues/416.

PR #35, opened 2020-07-22
Fix initial rendering

### LINKED ISSUE #32: No rendering without xochitl
Nothing gets rendered on the display if `xochitl` wasn't started at least once.

Steps to reproduce:

 1. Deploy demo application on device
 2. `ssh` to device
 3. `systemctl disable xochitl`
 4. `reboot`
 5. `ssh` to device
 6. `./demo` (you won't see the demo, but **it is** running, e.g. touch events are logged)
 7. Kill `./demo` (e.g. with `ctrl-c`)
 8. `xochitl`
 9. Kill `xochitl` (e.g. with `ctrl-c`)
 10. `./demo` (now you can also see the demo)

Any help on how to get started debugging this problem would be appreciated.

vmx:
When an application that is using libremarkable is started right after
booting the device, it doesn't display anything.

In order to fix this problem, `pixclock` is set to 6250. This will result
in the expected flashing of the display as when starting other applications.
Though the rendering will be garbled/corrupted.

In order to fix that, the fix screen info needs to be read after the
var screen info was set. The fix screen info returns the number of lines
based on the rotation of the screen. Right after starting the reMarkable
the screen is set to landscape mode. libremarkable sets it to portrait mode
via the var screen info, which changes the line length of the fix screen info.

I also adjusted the `xres` and `yres` in the code to reflect the values
after the `ioctl` was successfully sent. Using the flipped values would
still work as they are automatically corrected, but using the right values
makes the code easier to understand.

I've tested it with the `demo`, both the traditional build as well as with
the musl build.

Fixes #32.

### DIFF
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -43,14 +43,13 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
             .open(path_to_device)
             .unwrap();
 
-        let fix_screen_info = Framebuffer::get_fix_screeninfo(&device);
         let mut var_screen_info = Framebuffer::get_var_screeninfo(&device);
-        var_screen_info.xres = 1872;
-        var_screen_info.yres = 1404;
+        var_screen_info.xres = 1404;
+        var_screen_info.yres = 1872;
         var_screen_info.rotate = 1;
         var_screen_info.width = 0xffff_ffff;
         var_screen_info.height = 0xffff_ffff;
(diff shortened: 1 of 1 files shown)


vmx:
I've tried to keep the changes to a minimum. There are other values that a minimal example based on QT would set differently. Though as it works for me, I lean towards keeping the minimal changes in order not to possibly break older versions of the reMarkable.

canselcik:
@vmx thank you for getting to the bottom of this. Patch looks great. Merging. Much appreciated.

PR #38, opened 2020-08-18
Use Rust 2018 edition idioms

vmx:
This is a PR that updates the codebase to Rust 2018 editions idioms. It also does a few more cleanups I thought make sense. I wanted to make other change to libremarkable, but I'm so used to Rust 2018 edition codebases, hence this PR.

@canselcik please see this as a proposal. I intentionally tried to keep the commits small and self contained, so that you can decide if there are changes you don't like. I'm happy to rebase things and removing stuff or making other changes.

One issue is that `ioctl-gen` cannot be used with normal imports. I've created https://github.com/quadrupleslap-old/ioctl-gen/pull/1 to fix this. For now I've put my branch into Cargo.toml, that of course needs to be resolved before merging.

The demo application still works on my device.

PR #41, opened 2020-09-10
Rewrite multitouch (breaking changes!)

LinusCDE:
Hi,
here is the rewrite for multitouch.

- The `MultitouchEvent::Touch` event was split into `Press` and `Move`
- Added `MultitouchEvent::Release`
- Events provide a finger struct
- State machine for finger recognition was rewritten
  - (multitouch recognition of independent 10 fingers was successfully tested)

---

**Note**:

Since the implementation now uses a `Mutex<FxHashMap>>` instead of atomic values, it is techicially a bit slower. Since the performance hit is really really tiny, I haven't fixed it (though `FxHashMap` is already faster than `HashMap` of std). Possible solutions would be:
 - Use an array to completely abandon hashing (the slots should be predictable anyway)
 - Use something like [evmap](https://lib.rs/crates/evmap) which would eliminate any performance hit caused by the mutex as well (though having many writes may not be the ideal usecase for this crate)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,12 +43,12 @@ extern crate ioctl_gen;
 
 extern crate aabb_quadtree;
 extern crate atomic;
+extern crate fxhash;
 extern crate hlua;
 extern crate libc;
 extern crate mmap;
 extern crate rusttype;
 extern crate zstd;
-extern crate fxhash;
 
 pub extern crate cgmath;
 pub extern crate epoll;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,7 @@ extern crate libc;
 extern crate mmap;
 extern crate rusttype;
 extern crate zstd;
+extern crate fxhash;
 
 pub extern crate cgmath;
 pub extern crate epoll;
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -43,12 +43,12 @@ extern crate ioctl_gen;
 
 extern crate aabb_quadtree;
 extern crate atomic;
+extern crate fxhash;
```
LinusCDE, line 46: This will probably also have an impact on PR #38 . @vmx

fenollp:
10 fingers multitouch & tracking, that's nice!
IMO this should get in before #38 as its changes seem automatable enough.

LinusCDE:
I also created a small [swipe-engine](https://github.com/LinusCDE/retris/blob/master/src/swipe.rs) that was aimed a my [tetris clone implementation](https://github.com/LinusCDE/retris/) for the reMarkable but may also be handy for other users (const variables probably need to be changed a bit though and made configurable).  
I'm fine if you just take that file and integrate it with any needed/wanted changes into this lib.

vmx:
It's fine if this one merged before my #38, I can rebase. Though I'd prefer if I'd need to rebase only once. So it would be cool to merge things that should be merged first and then let me know when it's the turn for my PR so that I can then rebase it.

canselcik:
This is awesome and as you all agreed, I'm merging this first so that #38 doesn't need to be rebased multiple times. Excellent work.

PR #43, opened 2020-09-13
chore: add basic CI

vmx:
Provide basic CI. It makes sure that the code is formatted according
to rustfmt and that there are no clippy warnings.

Cross-compilation support could be added, but I thought as a starts, basic CI is better than none.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,65 @@
+on: [push, pull_request]
+
+name: Continuous integration
+
+jobs:
+  check:
+    name: Check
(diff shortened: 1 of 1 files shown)


vmx:
An example of the run of the GitHub Action can be found here: https://github.com/vmx/libremarkable/runs/1109258378

canselcik:
I've been meaning to get to setting up something like this for this repo for a while now but hadn't had the time. Thank you for doing this. As you mentioned, it would be even cooler if it did cross-compilation stuff. Definitely excited to merge those as well if you get a chance. :+1:

PR #44, opened 2020-09-21
Fix thread panic when resumed in remux

LinusCDE:
When paused (SIGSTOP) and resumed (SIGCONT) in remux, the ev eventloop would panic on the replaced unwrap() call, which led to missing input events.

This was brought up in issue https://github.com/LinusCDE/retris/issues/1. See there for more information.


canselcik (approved):
Thank you, LGTM.

PR #47, opened 2020-11-01
Fix gpio and add input example

LinusCDE:
Hi,

I'm currently on adding rm2 support to the input system. At first I added a `examples/input.rs` to debug differences and noticed that the `GPIOEvent::Unpress` never happened.

I decided to make this a separate PR since it doesn't add any rM2 updates, yet. It does 2 things however:
- Add the Debug trait to all InputEvent related structs
- Add an `input` example which simply prints the debug information of all input events (similar to evtest)
- Remove the GPIO debouncing to get the Unpress-Event to work

(This PR will probably be followed by one for the rM2 support shortly.)

### DIFF
--- a/examples/input.rs
+++ b/examples/input.rs
@@ -1,0 +1,16 @@
+use libremarkable::input::{ev::EvDevContext, InputDevice, InputEvent};
+use std::sync::mpsc::channel;
+
+fn main() {
+    // Send all input events to input_rx
+    let (input_tx, input_rx) = channel::<InputEvent>();
+    EvDevContext::new(InputDevice::GPIO, input_tx.clone()).start();
+    EvDevContext::new(InputDevice::Multitouch, input_tx.clone()).start();
+    EvDevContext::new(InputDevice::Wacom, input_tx).start();
+
+    eprintln!("Waiting for input events...");
+    while let Ok(event) = input_rx.recv() {
+        println!("{:?}", event);
+    }
+    eprintln!("All event loops were closed?!?");
+}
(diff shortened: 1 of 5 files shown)


LinusCDE:
As a demonstration, here are the input examples before commit cca4397 (`input-prefix`) and after it (`input-postfix`): [input-examples.zip](https://github.com/canselcik/libremarkable/files/5471178/input-examples.zip)

I ran both on my rM1 and pressed the Middle Button twice:
```
reMarkable: /home/root/ ./input-prefix 
Waiting for input events...
GPIO { event: Press { button: MIDDLE } }
GPIO { event: Press { button: MIDDLE } }
^C
reMarkable: /home/root/ ./input-postfix
Waiting for input events...
GPIO { event: Press { button: MIDDLE } }
GPIO { event: Unpress { button: MIDDLE } }
GPIO { event: Press { button: MIDDLE } }
GPIO { event: Unpress { button: MIDDLE } }
^C
reMarkable: /home/root/ 
```

canselcik:
Looks like #48 had most of the changes that would have come from this patch except for a small diff on `examples/input.rs`. Let me know if you'd like to rebase and merge.

LinusCDE:
Just tested the current input example binary from the master branch and it seems that I integrated the fix in the previously merged pr. I tried to get the changes apart at the beginning, but the fact that I needed the input example to have proper input tests made it hard.

I'll close this as the changes seem to be already in. Maybe I mentioned that in the ther PR.

PR #49, opened 2020-11-28
CI: fix typo that blocks demo autorelease

fenollp:


LinusCDE:
Not tracking the `Cargo.lock` seems to be the intended practice for rust libraries (but for bins) if I'm not mistaken.

The changes seem totally fine to me.

PR #51, opened 2020-12-24
Add support for the rM2 battery

LinusCDE:
Commit message: This was not tested. Though by viewing the file tree of /sys provided by @torwag [here](https://github.com/Eeems/oxide/issues/48#issue698181952) (file tree.txt line 3166) it seems that all methods are supported.

I didn't test whether this actually works on a rM 2 as I don't own that device. I'm confident though that it should work. I added the compiled demo binary that also reports that battery data. Please test whether it displays Charging/Discharging and the percentage on the reMarkable 2.

---

- [demo.tar.gz](https://github.com/canselcik/libremarkable/files/5740678/demo.tar.gz)

### DIFF
--- a/src/battery.rs
+++ b/src/battery.rs
@@ -1,11 +1,19 @@
+use crate::device::CURRENT_DEVICE;
 use std::fs::File;
 use std::io::Read;
 
 // TODO: Implement API to allow callbacks backed via uevent / inotify
 
+// File tree containing the rM2 battery:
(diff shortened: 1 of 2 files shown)


raisjn:
looks good to me, shows 'Full - 99%'. I don't see anything about charging / discharging

LinusCDE:
Thanks for testing! Seems to work then. I guess that "Full" is a third state I didn't know about meaning "would charge if battery allowed for it". It's just the file content of the sysfs battery interface.

This PR is then mergable in my opinion since it is now tested on an actual rM 2.

PR #53, opened 2021-01-01
fix: Skip groupadd if gid exists in Docker setup

vidbina:
Users who have a gid that matching an existing gid in the container will encounter an error when running `groupadd`. In order to allow the docker build process to continue, the addition of the `-f` flag will allow `groupadd` to exit successfully if the group already exists.

### DIFF
--- a/docker-toolchain/Dockerfile
+++ b/docker-toolchain/Dockerfile
@@ -11,7 +11,7 @@ ARG ostype=Linux
 RUN apt-get -qq update
 RUN apt-get -qq install curl build-essential gcc-arm-linux-gnueabihf vim
 
-RUN bash -c 'if [ ${ostype} == Linux ]; then groupadd -r --gid ${GID} ${UNAME}; fi'
+RUN bash -c 'if [ ${ostype} == Linux ]; then groupadd -f -r --gid ${GID} ${UNAME}; fi'
 RUN useradd -u $UID -g $GID -m $UNAME
 USER $UNAME
 


SafariMonkey:
I'm not convinced that this is the best solution. I'd like to point out the following sentence from the documentation:

> When used with -g, and the specified GID already exists, another (unique) GID is chosen (i.e. -g is turned off).

This means that it will create a group with that name but a different GID unnecessarily.

If the GID exists, the next line will run without an issue. If it doesn't, it'll presumably fail, based on the note:

> A group number must refer to an already existing group.

Based on this, I suggest, as a possibility, just adding `|| true` to the previous line. The command output will still show in the console if it fails, and if the next line succeeds, all is well.

`-f` should be fine, of course, as the extra group is only being created inside the container. This just seems like a cleaner solution to me.

vidbina:
Good spot, @SafariMonkey. Based it from the `groupadd --help` in the base image which mentioned the following for  `-f`:

> exit successfully if the group already exists, and cancel -g if the GID is already used

I understood this bit to mean that the use of `-f` with `-g` would be "safe enough".

```
└──> docker run --rm -it ubuntu:16.04 bash
root@8e4d4c8550cd:/# groupadd --help
Usage: groupadd [options] GROUP

Options:
  -f, --force                   exit successfully if the group already exists,
                                and cancel -g if the GID is already used
  -g, --gid GID                 use GID for the new group
  -h, --help                    display this help message and exit
  -K, --key KEY=VALUE           override /etc/login.defs defaults
  -o, --non-unique              allow to create groups with duplicate
                                (non-unique) GID
  -p, --password PASSWORD       use this encrypted password for the new group
  -r, --system                  create a system account
  -R, --root CHROOT_DIR         directory to chroot into
      --extrausers              Use the extra users database
```

Did look into the resulting image and notice that the `builder` group was created with a new gid (999 in my case), so I have updated the PR such that we don't have this rogue group in the the image. Thanks for your input :pray: 

@canselcik let me know if this is worthwhile merging or if you'd like me to supply this PR squashed.

PR #54, opened 2021-01-02
Replace Model::Unknown with UnknownVersion error

rien:
When the `current_model` is `Unknown`, it is unlikely implementors will be able to do meaningful things with it. In my opinion is makes more sense to move this flow to an error result.

I have created an `ErrorKind` which can either be an `IOError` and an `UnknownVersion` so it is clear why determination of the current model failed.

This has two main advantages:
- The code for handling different device models is simpler: only "happy path"'s need to be considered.  The pattern match `Model::Unknown => unreachable!()` is no longer needed.
- The `UnknownVersion`  error now shows what the version string is that wasn't recognized. This will speed up adding the version to the known versions if the user reporting the error includes their error.

The disadvantage is however that the return type of `current_model()` and the `Model` enum have changed in a way that breaks the current API, thus requiring a major version bump. I do think this change improves usability of the API, but I would understand if you want to reject or postpone this for API stability.


SafariMonkey:
I have some breaking change ideas of my own. I'd like to do a write-up of my ideas and maybe a draft PR, and submit them in the next few days. They're significantly larger, though.

Just thought I'd mention it since we're discussing a breaking version bump. The change here doesn't block anything other than nicer ergonomics, so it may be worth waiting to see if they can be bundled (if my suggestions are welcomed).

LinusCDE:
Good ideas. I also don't think that breaking changes will be any problem, I I'm most likely the only one having ever used this still quite new api.

LinusCDE:
@rien Also, when you're already on it, it would be nice to have `current_version()` be pub. That way software can easily determine if the version is unknown and do custom handling with it. Otherwise no-one could use the new great Error anyway and having it being public wouldn't make sense.

rien:
@LinusCDE I have made `current_model()` public. For a library it is indeed more logical to allow implementors of the library choose to handle the error, instead of `panic!`ing. Is it maybe an idea to make `CURRENT_DEVICE` a `Result` instead of panicking?

LinusCDE:
> @LinusCDE I have made `current_model()` public. For a library it is indeed more logical to allow implementors of the library choose to handle the error, instead of `panic!`ing. Is it maybe an idea to make `CURRENT_DEVICE` a `Result` instead of panicking?

I think this is good as it is. The `CURRENT_DEVICE` could be a result, but it is probably more expected to *just be there*.

SafariMonkey:
I have made some attempts at the breaking changes I think would be useful, but the one I was pursuing has turned out to be more _annoying_ as a change than I'm really happy with. It requires very pervasive injection of generics in a way where I'm uncertain as to whether it's worth it.

I will try to put up draft PRs tonight, with incomplete sketches of implementations - I expect them to be more of a discussion point than something to polish and merge. One is pretty nice and I've been using it in my experiments. The other one is an extension of the first concept that just ended up being too invasive, and I still don't have an implementation of it I'd say I'm happy with. Maybe it's something that could be handled with a redesign, but at that point we're talking about more architectural changes.

PR #62, opened 2021-03-28
Update some info in the readme

bkirwi:
As discussed in Dischord!

I dropped the entire musl section for now - we can always bring it back if we find instructions that work, but for now having it in seems confusing.

Also added a couple minor tweaks, like pointing to the wiki for the toolchain while that's in limbo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,17 @@ make sure to include `--release --target=armv7-unknown-linux-gnueabihf` in your 
     Finished dev [unoptimized + debuginfo] target(s) in 24.85 secs
 ```
 The `--release` argument is important as this enables optimizations and without optimizations you'll be looking at ~70% CPU utilization even when idle. With optimizations, the framework runs really light, 0% CPU utilization when idle and 1-2% at peak.
+
+#### Building demo with `cross`
+[`cross`](https://github.com/rust-embedded/cross) is a utility for cross compilation and “cross testing” of Rust crates.
+It allows building `libremarkable` apps without the official toolchain.
+1. Install `cross` with `cargo install cross` (make sure the Remarkable toolchain is not in use first)
+1. Compile with `cross build --example demo --release --target=armv7-unknown-linux-gnueabihf`
+
+Note:
+* Building this way does not require Remarkable's toolchain nor building on Ubuntu 16.04 so setting up should be easier.
+* It's also possible to build against [`musl`](https://musl.libc.org/) a minimal `libc` implementation:
+  just substitute `armv7-unknown-linux-musleabihf` in the instructions above.
+  However, **this will render the framebuffer unusable on reMarkable 2**:
+  [the framebuffer shim](https://github.com/ddvk/remarkable2-framebuffer) does not appear to work correctly with `musl`.
+  * Make sure to build with `lto = true` otherwise `musl` symbols may be improperly resolved (call to `mmap` fails).
--- a/README.md
+++ b/README.md
@@ -19,7 +19,10 @@ For further documentation see the [wiki](https://github.com/canselcik/libremarka
 ### Build Instructions
 
 #### Setting up the toolchain
-In order to build `libremarkable` and the examples (`spy.so` and `demo`), you'll need the toolchain from Remarkable. Download the [installation script](https://remarkable.engineering/oecore-x86_64-cortexa9hf-neon-toolchain-zero-gravitas-1.8-23.9.2019.sh) from [remarkable.engineering](https://remarkable.engineering/) and install the toolchain.
+In order to build `libremarkable` and the examples (`spy.so` and `demo`),
+you'll need the toolchain from Remarkable.
+Check [the community wiki](https://remarkablewiki.com/devel/toolchain)
+for the latest instructions on where to find it.
 
 You can then set up your Rust toolchain for cross compilation with: `rustup target add armv7-unknown-linux-gnueabihf`.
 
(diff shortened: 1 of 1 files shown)


fenollp:
I think it can be useful to keep the section on `cross` just replacing mentions of musl with gnueabihf. This works in the CI today and does not necessitates the oecore toolchain nor special linker flags outside lto=true. 

Not sure what's wrong with musl though!

What's the discussion you're mentioning? On which Discord?

bkirwi:
Apologies, here's the thread: https://discord.com/channels/385916768696139794/386181213699702786/824072591881797662

I'd be happy to have `cross`-based instructions, but I never quite got the whole thing to work, so I don't feel super qualified to write them! In the meantime it seemed useful to remove some content that was tripping people up.

My impression is that statically linking `musl` interacts badly with the framebuffer shim, which relies on being able to intercept libc calls that don't happen when you're not dynamically linking libc. @LinusCDE also mentioned some recent issue with `evdev`, though I'm not familiar.

fenollp:
Alright I fixed building with musl (works for rM device v1) in https://github.com/canselcik/libremarkable/pull/64 and hinted a bit more on `cross` usage.
Thanks for pointing out the problem I think this is now superseeded.

Regarding `musl` / the rm2fb shim / `LD_PRELOAD` / the rM2, it *may work (quoting [the musl FAQ](https://wiki.musl-libc.org/faq.html))* but I have not tested this (yet anyway):
> The environment variables LD_PRELOAD and LD_LIBRARY_PATH are also supported for one-off cases.

LinusCDE:
Just gave @fenollp's [musl build](https://github.com/fenollp/libremarkable/tree/musl) a try. The shim doesn't work on it indeed since no interpreter (usually some linux-ld library) is declared at all and this preventing any ld based interception. The above poc could for example only be enabled on musl builds to make them still work with the shim.

fenollp:
> Feel free to do whatever you want with this poc. I personally don't plan on continuing on it in the foreseeable future and don't care what happens with that at all.

Hey thanks for this. Is everything alright with you? your language seems kind of harsh.

This is a pretty sic PoC if you ask me :) I'll definitely continue your work.
Thoughts: rm2fb's `close` can maybe be called from `libc::atexit` and surely the two `ioctl_func` symbols can be resolved during lazy init or cached in some way.

LinusCDE:
> Hey thanks for this. Is everything alright with you? your language seems kind of harsh.

Sorry for that. Had a long day and wanted to make it clear that there is no issue in coping/using it at all. I'm sorry if that came forth as harsh. I didn't proof-read that message and it seems harsh indeed. 

> Thoughts: rm2fb's `close` can maybe be called from `libc::atexit` and surely the two `ioctl_func` symbols can be resolved during lazy init or cached in some way.

I gave lazily getting the `ioctl_func` a shot and ran into some weird issue that this specific static ref wasn't declared.

For the close I was more concerned that rust will issue a normal close once Framebuffer gets dropped. If a library would for some reason want to reopen it afterwards, rm2fb wouldn't have expected the fd to be closed. Implementing `Drop` for the framebuffer and doing a `self.device.into_raw_fd()` should probably prevent the file from getting closed.

A quick test also turned out that this won't work easily for a musl build. My guess is that libloader uses ld itself. Should we be able to add libloaders dependencies statically I'm not sure how this would be for loading rm2fb. It would probably require to at least load libc again anyway. Maybe creating a rm2fb rust crate with the option to use either a dynamically linked library (rust *-sys crate) or vendoring it would be useful for musl. I thinking of how [openssl](https://lib.rs/openssl) is done. You can specify the "vendored" feature to get it compiled into the rust binary.

As of now the poc can only serve to remove the need of using LD_PRELOAD. Not sure how useful that is.

bkirwi:
That is definitely a very cool direction... seems like it would make adopting the library and cross-building for RM1 & RM2 much simpler. (From my perspective the main appeal of the `musl` instructions was the use of `cross` / not needing the "official" toolchain which is increasingly difficult to access. So I'd personally be excited by this approach even if it didn't improve the `musl` issue.)

Do folks think this PR is worth landing in the meantime? Or should I close and open a new issue for the `LD_PRELOAD`-less approach?

LinusCDE:
> Do folks think this PR is worth landing in the meantime? Or should I close and open a new issue for the LD_PRELOAD-less approach?

Yeah. I think that the `LD_PRELOAD`-less approach should be a different PR. For now having musl support is already a good thing. Even if this makes the framebuffer for the rM 2 useless in the musl-build. Maybe a mention regarding this problem could be added to the readme as well.

fenollp:
> From my perspective the main appeal of the musl instructions was the use of cross / not needing the "official" toolchain which is increasingly difficult to access

This is possible today (no musl involved). Works for rM1 & rM2:
```
cross build --example demo --release --target=armv7-unknown-linux-gnueabihf
```
This only requires `cargo install cross` (and Docker).

---

> Maybe a mention regarding this problem could be added to the readme as well.

Mentioned in my PR. Thanks

bkirwi:
> This is possible today (no musl involved)

Yep, figured this out eventually... but it wasn't clear when I was first learning this stuff a couple months back!

> Even if this makes the framebuffer for the rM 2 useless in the musl-build. Maybe a mention regarding this problem could be added to the readme as well.

I restored the section on `cross`, but changed it to mention the glibc version first; it still mentions `musl`, but also that it's not working correctly on rm2 at the moment.

Maintainers should be allowed to edit this branch, so if there are tweaks to the wording you prefer feel free to make them directly!

bkirwi:
Ah, just saw https://github.com/canselcik/libremarkable/pull/64/files, which covers this and more... will close this one down. Thanks for the discussion!

bkirwi:
Opened #65 as a tracking issue for the LD_PRELOAD-less approach!

Issue #57, opened 2021-03-18
Compilation Error with EvDev (Mismatch types)


rlingineni:
I was following the instructions to build with `musl`.

However, there seems to be a compile error. Building on MacOS.

```
Compiling evdev v0.10.2
convert_ioctl_res!(::nix::libc::ioctl(fd, ior!(b'E', 0x20 + ev, len) as ::libc::c_ulong, buf))
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
```
I'm new to the lang, but it seems the error might be due to this issue: https://github.com/emberian/evdev/issues/22 

Looks like the newest version hasn't been published with the fix

bkirwi:
Fixed now that the linked PR is merged... thanks for the report!

PR #64, opened 2021-03-29
Describe how to build without reMarkable toolchain using `cross` and fix building with `musl`

fenollp:
Fixes https://github.com/canselcik/libremarkable/issues/57


bkirwi:
Merged as part of #68!

Issue #65, opened 2021-04-01
Avoid needing LD_PRELOAD on reMarkable 2


bkirwi:
Originally discussed here: https://github.com/canselcik/libremarkable/pull/62#issuecomment-809655455

The basic idea is to issue calls to the `/opt/lib/librm2fb_client.so.1` library directly, if it's present. This means `libremarkable` apps on rm2 would "just work" without needing to use the shim explicitly. (Though it would still need to be installed.)

LinusCDE:
Here is a working (but terribly slow) poc: https://github.com/LinusCDE/libremarkable/tree/explicit_rm2fb_poc

Anyone can feel to expand on that.

bkirwi:
A second approach has been implemented at https://github.com/canselcik/libremarkable/pull/78 - implementing the rm2fb client logic in Rust instead of depending on `librm2fb_client` at all.

PR #76, opened 2021-10-22
Support rM2 for musl builds

fenollp:
Fixes https://github.com/canselcik/libremarkable/issues/65
Requires https://github.com/ddvk/remarkable2-framebuffer/pull/78

Loads librm2fb client lib on startup and uses it directly in lieu of going through libc (with LD_PRELOAD).
This is the only way to run binaries that use musl (instead of libc) on rM2.
Using musl isn't common in the homebrew rM apps community however. Non-musl binaries work fine with Toltec's tools on rM2. And rM1 has no issue as it doesn't have a fake framebuffer.

So far it seems on par with the LD_PRELOAD solution.
I haven't looked at precise measurements of speed nor CPU load. Running the demo seems to be as responsive when drawing with the pencil and multiple fingers. UI redraw is painfully slow (bottom buttons) and takes about 10s to complete, however that is the case with both solutions.

So I am suggesting to have this way of using rm2fb be enabled only for musl targets.
I haven't done that part yet though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ atomic = { version = "0.5.0" }
 cgmath = "0.17.0"
 fxhash = "0.2.1"
 once_cell = "1.7.2"
+libloading = "0.7.1"
 
 [features]
 enable-runtime-benchmarking = []
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -1,3 +1,4 @@
+use log::debug;
 use memmap2::{MmapOptions, MmapRaw};
 
 use std::fs::{File, OpenOptions};
@@ -35,26 +36,19 @@ unsafe impl<'a> Sync for Framebuffer<'a> {}
 
 impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
     fn from_path(path_to_device: &str) -> Framebuffer<'_> {
-        let device = if true {
-            if let Some(lib) = &*LIBRM2FB_CLIENT {
-                unsafe {
-                    let c_str = std::ffi::CString::new(path_to_device).unwrap();
-                    let c_world: *const libc::c_char = c_str.as_ptr() as *const libc::c_char;
-                    let open_func: libloading::Symbol<
-                        unsafe extern "C" fn(
-                            fd: *const libc::c_char,
-                            flags: libc::c_int,
-                            mode: libc::mode_t,
-                        ) -> libc::c_int,
-                    > = lib.get(b"open").unwrap();
-                    let fd = open_func(c_world, libc::O_RDWR, 0 as libc::mode_t);
-                    println!("FD is {}", fd);
-                    std::fs::File::from_raw_fd(fd)
-                }
-            } else {
-                panic!("Failed to open file with shim lib!");
+        let device = if let Some((open_func, _, _)) = &*LIBRM2FB_CLIENT {
+            unsafe {
+                let c_str = std::ffi::CString::new(path_to_device).expect("CString::new failed");
+                let fd = open_func(
+                    c_str.as_ptr() as *const libc::c_char,
+                    libc::O_RDWR,
+                    0 as libc::mode_t,
+                );
+                debug!("[rm2fb] open: using client, FD is {}", fd);
+                std::fs::File::from_raw_fd(fd)
             }
         } else {
+            debug!("[rm2fb] open: using libc");
             OpenOptions::new()
                 .read(true)
                 .write(true)
@@ -83,7 +77,6 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
 
         let fix_screen_info = Framebuffer::get_fix_screeninfo(&device);
         let frame_length = (fix_screen_info.line_length * var_screen_info.yres) as usize;
-        println!("FIX: {:#?}\nVAR: {:#?}", fix_screen_info, var_screen_info);
         let mem_map = MmapOptions::new()
             .len(frame_length)
             .map_raw(&device)
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -42,7 +42,7 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
                 let fd = open_func(
                     c_str.as_ptr() as *const libc::c_char,
                     libc::O_RDWR,
-                    0 as libc::mode_t,
+                    0_u32, /*as libc::mode_t*/
                 );
                 debug!("[rm2fb] open: using client, FD is {}", fd);
                 std::fs::File::from_raw_fd(fd)
@@ -77,6 +77,7 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
 
         let fix_screen_info = Framebuffer::get_fix_screeninfo(&device);
         let frame_length = (fix_screen_info.line_length * var_screen_info.yres) as usize;
+
         let mem_map = MmapOptions::new()
             .len(frame_length)
             .map_raw(&device)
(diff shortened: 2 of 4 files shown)


LinusCDE:
> @LinusCDE `cross build -vv --examples --target=armv7-unknown-linux-gnueabihf` seems to solve the architecture incompatibilities, however I get stuck with `ld` looking for libstdc++ symbols.
> 
> Note: this doesn't seem to help `echo 'QMAKE_LFLAGS += -static -static-libgcc -static-libstdc++' >>src/client/client.pro`

Adding those deps seems to resolve this:

```toml
[dependencies]
cpp = "0.5"

[build-dependencies]
cpp_build = "0.5"
```

However I have not gotten stuck on adding QMake stuff. Prebuilding it was my hack for now. But still need to add libfrida-gum (which is a static lib in the repo). Building it [from deps](https://github.com/LinusCDE/rm2fb_client-rs/blob/more_foolery/Cargo.toml) seems to have some issues because it's forcing the `-m64` flag and has reqwest which requires a vendored openssl dep but somehow fails that. [Adding the existing lib](https://github.com/LinusCDE/rm2fb_client-rs/blob/32b83e17322ebc6439f0aab6a2afee807378f348/build.rs#L69) seems the most sensible but libc doesn't include `fnctl64` (linking error) and I somehow can't manually declare that it should exist.

fenollp:
I got around cpp linking with this line from https://flames-of-code.netlify.app/blog/rust-and-cmake-cplusplus/
```rust
println!("cargo:rustc-link-lib=dylib=stdc++");
```
Not sure how different it is from relying on cpp/cpp_build dependencies. Surely it still relies on the system-wide C++ compiler?

Adding `frida-gum` the same way `rm2fb_client` is linked works, and that leaves me with the same linking issues you have: qt+fcntl64.


All that said at this point I'd rather attempt a Rust implementation of `rm2fb_client`.

LinusCDE:
> I got around cpp linking with this line from https://flames-of-code.netlify.app/blog/rust-and-cmake-cplusplus/
> 
> ```rust
>  println!("cargo:rustc-link-lib=dylib=stdc++");
> ```

Cool. I attempted something similar, but failed since I tried only static linking.

> All that said at this point I'd rather attempt a Rust implementation of rm2fb_client.

Couldn't agree more. Seems portable anyway. We just need to properly interact with `/dev/shm/swtfb.01` if I'm not mistaken.

fenollp:
Indeed, we just need to implement this here https://github.com/ddvk/remarkable2-framebuffer/issues/11#:~:text=how%20to%20write%20an%20IPC%20client%20for%20rm2fb
Maybe with https://crates.io/crates/ipc-channel which seems to be using https://crates.io/crates/bincode for serialization.
Or with https://crates.io/crates/interprocess which seems more actively maintained.	

Actually https://docs.rs/msg/0.1.1/src/msg/macros.rs.html#43-63 should do for `msgget`/`msgsnd`. I am just unsure about the serialization part...

LinusCDE:
Did some attempt at reimplementing the client. It should work, but I fear that I have some issue with the shared buffer rn. On first try it can send a couple of updates but it gets stuck at some point and according to the live example with shim, the buffer didn't change as well. (You need to restart your device to get it unstuck). I also confirmed that the size of the swtfb_update struct is the same size in rust and the original cpp client.

Can you take a look at the feature/integrate_swtfb_client branch @fenollp ?

If that works, we still need to sort out how to decide what to use. Especially since we might need to interfere with the chosen path in `from_path(&str)` to use the correct shared memory. There are also some unimplemented ioctls. Those shouldn't be a problem though since they're probably all noops.

---

Update: Now I get a "Bus Error". If definitely fucked up something with the memory map. Not sure what. Maybe I'll take a look in a couple of days again. Feel free to tinker around.

LinusCDE:
The framebuffer is now fixed with commit cf8d742b. Since I last fixed the wrong type of `mtype` it seems that at least messages are now sent without any issue.

So the framebuffer can now be changed on `/dev/shm/swtfb.01` (confirmed with the live example (old build with shim)), but the updates are now applied. At least the queue doesn't get stuck anymore. So just some problem regarding the actual data I guess.

(Btw I didn't use the interprocess lib since it doesn't seem to support SysV Message Queues. There is only one lib supporting it (ipc-rs) but it seems abandoned and doesn't even have a git repo listed). So libc seems like the best bet here.

LinusCDE:
IT WORKS!!!

I intercepted the old demo and the new one and found the mistake: I sent a pointer of a reference/pointer! Effing stupid part on my site. Will do some polish, but it works already!!!

LinusCDE:
Musl build runs fine:

Musl build of demo: [demo.zip](https://github.com/canselcik/libremarkable/files/7532314/demo.zip)

Note: Some functions where changed, but those are probably not used much. We may want to refactor the framebuffer into a "Core" and "Rm2fb" backend and use this opportunity to refactor the framebuffer stuff in general.

But all in all this should work now. A user needs to point the framebuffer to `/dev/shm/swtfb.01` to make use of the internal client. Since the shim only hooks ioctls and opening `/dev/fb0` the own rm2fb client doesn't interfere with the `LD_PRELOAD`ed lib. It will just not get used (maybe it'll have an opened framebuffer lying around).

bkirwi:
bq. We may want to refactor the framebuffer into a "Core" and "Rm2fb" backend and use this opportunity to refactor the framebuffer stuff in general.

Agree it would be nice to restructure here a bit. I've also thought about adding an in-memory backend for the framebuffer, mainly for testing... I suspect any refactoring that split the core and rm2fb stuff would extend nicely to handle that case too.

bq. Since the shim only hooks ioctls and opening /dev/fb0 the own rm2fb client doesn't interfere with the LD_PRELOADed lib. It will just not get used (maybe it'll have an opened framebuffer lying around).

Makes for a nice migration path!

LinusCDE:
> > We may want to refactor the framebuffer into a "Core" and "Rm2fb" backend and use this opportunity to refactor the framebuffer stuff in general.
> 
> Agree it would be nice to restructure here a bit. I've also thought about adding an in-memory backend for the framebuffer, mainly for testing... I suspect any refactoring that split the core and rm2fb stuff would extend nicely to handle that case too.

I'm not sure what the performance implication on batch `set_pixels` would be if we changed the fields `device`, `mem_map` and `swtfb_client` into a enum with two backends.

Other solution would be to turn `core.rs` into a package (`core/mod.rs`) which should keep the semantics the same and create a `SwtFramebuffer` that implements traits like `core::Framebuffer` does. That way we could swap how memory is accessed easily and it might retain the exact same performance for core. Only issue would be reimplementing traits that are independent. But i think implementing `FramebufferDraw` ontop of `dyn FramebufferIO` should work in this case and reduce the actual amount of traits to be implemented. Maybe then offer a builder-like function to replace the explicitly to `core::Framebuffer` bound function `from_path()`.

I think Both approaches would allow for a third `test` backend. Whether a Memmap is just allocated in heap (for 1st aproach) or we re-implement `FramebufferIO` do do something entirely else is not that big of a different in effort and should amount to roughly the same benefit.

> > Since the shim only hooks ioctls and opening /dev/fb0 the own rm2fb client doesn't interfere with the LD_PRELOADed lib. It will just not get used (maybe it'll have an opened framebuffer lying around).
> 
> Makes for a nice migration path!

Yeah. Who doesn't care will in 99% percent of case not migrate automatically and also not experience any of the breaking changes as well. Perfect if you just want to bump the version but don't care about what changed from a technical point of view.

LinusCDE:
One idea for refactoring would probably also be using something like [thiserror](https://lib.rs/thiserror) for custom errors (instead of except()s or just relaying IoErrors) This is probably a bigger undertaking und more fit for a general refractor/unifying of some names.

For starters adding a easily findable method to determine whether swtfb.01 is supported (explicit check for the file or checking for the existance of the SysV bus) would be better so an App can error out more nicely without needing to get deep into the specifics to know why.

fenollp:
Closing in favor of https://github.com/canselcik/libremarkable/pull/78

PR #78, opened 2021-11-13
Add built-in swtfb client and enable musl builds on rM 2

LinusCDE:
See #76 . This should also enable musl builds and hence solve that problem.

Code should be semi clean. As mentioned in #76 we might want to change "Core" and "Swtfb" into seperate backends (also todo when doing breaking changes: use `AsRef<std::path::Path>` in `from_path`).

Testing if wanted:
 - Musleabihf build should work for demo example
 - Demo example with libc should run with AND without shim (e.g. LD_PRELOAD=... or rm2fb-shim).

Artifacts built with `cross build --target=armv7-unknown-linux-gnueabihf --release --example demo` and `cross build --target=armv7-unknown-linux-musleabihf --release --example demo` : [demos.tar.gz](https://github.com/canselcik/libremarkable/files/7532384/demos.tar.gz)

New:
 - `framebuffer::swtfb_client`
 - `examples/swftb_sysv_spy.rs` as dylib (for LD_PRELOAD)
 - `device::Device::get_framebuffer_path(&self) -> &'static str`: Recommended to use in never versions as arg for `Framebuffer::from_path()`. Otherwise new swtfb_client will not be used automatically.

Breaking changes:
 - `get_fix_screeninfo`, `get_var_screeninfo`, `put_var_screeninfo` have got an additional parameter to work with swtfb_client. I doubt that anyone used such low level functions apart from me in plato.

Other changes:
 - Removed some duplicate code in `FramebufferRefresh`
 - Added new field `swtfb_client: Option<super::swtfb_client::SwtfbClient>` to `Framebuffer`

Note: With swtfb_client, FixScreeninfo->smem_start will not be set (= 0) since it's hard to do without splitting the `from_path` completely for both variants. Again I highly doubt that anyone used this. `Framebuffer.frame.as_ptr()` should return the correct value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,10 @@ all: library examples
 
 .PHONY: examples
 examples:
-	cargo build --examples --release --target=$(TARGET)
+	cargo build --examples --release --target=armv7-unknown-linux-gnueabihf
 
 demo:
-	cargo build --example demo --release --target=$(TARGET)
+	cargo build --example demo --release --target=armv7-unknown-linux-gnueabihf
 
 x-demo:
 	cross build --example demo --release --target=$(TARGET)
@@ -22,7 +22,7 @@ deploy-x-demo: x-demo
 	ssh $(DEVICE_HOST) 'RUST_BACKTRACE=1 RUST_LOG=debug ./demo'
 
 bench:
-	cargo build --examples --release --target=$(TARGET) --features "enable-runtime-benchmarking"
+	cargo build --examples --release --target=armv7-unknown-linux-gnueabihf --features "enable-runtime-benchmarking"
 
 .PHONY: docker-env
 docker-env:
@@ -42,10 +42,10 @@ examples-docker: docker-env
 		-v cargo-registry:/home/builder/.cargo/registry \
 		-w /home/builder/libremarkable \
 		rust-build-remarkable:latest \
-		cargo build --examples --release --target=$(TARGET)
+		cargo build --examples --release --target=armv7-unknown-linux-gnueabihf
 
 library:
-	cargo build --release --target=$(TARGET)
+	cargo build --release --target=armv7-unknown-linux-gnueabihf
 
 test:
 	# Notice we aren't using the armv7 target here
--- a/examples/swtfb_sysv_spy.rs
+++ b/examples/swtfb_sysv_spy.rs
@@ -1,7 +1,7 @@
 use libc::{c_int, c_void, key_t, msqid_ds, size_t};
 use redhook::{hook, real};
 
-use libremarkable::framebuffer::swtfb_ipc;
+use libremarkable::framebuffer::swtfb_client;
 
 hook! {
   unsafe fn msgget(key: key_t, msgflg: c_int) -> c_int => msgget_spy {
@@ -23,26 +23,26 @@ hook! {
     unsafe fn msgsnd(msqid: c_int, msgp: *const c_void, msgsz: size_t, msgflg: c_int) -> c_int => msgsnd_spy {
         let res = real!(msgsnd)(msqid, msgp, msgsz, msgflg);
         eprintln!("Spy: msgsnd({:#x}, {:?}, {}, {:#x}) => {}", msqid, msgp, msgsz, msgflg, res);
-        if msgsz == std::mem::size_of::<swtfb_ipc::swtfb_update>() {
-            let msg = &*(msgp as *const swtfb_ipc::swtfb_update);
+        if msgsz == std::mem::size_of::<swtfb_client::swtfb_update>() {
+            let msg = &*(msgp as *const swtfb_client::swtfb_update);
                 eprintln!("Spy: msgsnd: Message: swt_update.mtype: {:?}, data: ... }}", msg.mtype);
                 let data_str_formatted = match msg.mtype {
-                    swtfb_ipc::MSG_TYPE::INIT_t => {
+                    swtfb_client::MSG_TYPE::INIT_t => {
                         format!("...")
                     },
-                    swtfb_ipc::MSG_TYPE::UPDATE_t => {
+                    swtfb_client::MSG_TYPE::UPDATE_t => {
                         format!("{:?}", msg.data.update)
                     },
-                    swtfb_ipc::MSG_TYPE::XO_t => {
+                    swtfb_client::MSG_TYPE::XO_t => {
                         format!("{:?}", msg.data.xochitl_update)
                     },
-                    swtfb_ipc::MSG_TYPE::WAIT_t => {
+                    swtfb_client::MSG_TYPE::WAIT_t => {
                         format!("{:?}", msg.data.wait_update)
                     }
                 };
                 eprintln!("Spy: msgsnd: Message: swt_update.data: {} }}", data_str_formatted);
         }else {
-            eprintln!("Spy: msgsnd: Error: Message is not sizeof(swtfb_update) (expected {}, got {})", std::mem::size_of::<swtfb_ipc::swtfb_update>(), msgsz)
+            eprintln!("Spy: msgsnd: Error: Message is not sizeof(swtfb_update) (expected {}, got {})", std::mem::size_of::<swtfb_client::swtfb_update>(), msgsz)
         }
         res
   }
--- a/examples/swtfb_sysv_spy.rs
+++ b/examples/swtfb_sysv_spy.rs
@@ -1,0 +1,49 @@
+use libc::{c_int, c_void, key_t, msqid_ds, size_t};
+use redhook::{hook, real};
+
+use libremarkable::framebuffer::swtfb_ipc;
+
+hook! {
+  unsafe fn msgget(key: key_t, msgflg: c_int) -> c_int => msgget_spy {
+        let res = real!(msgget)(key, msgflg);
+        eprintln!("Spy: msgget({:#x}, {:#x}) => {:#x}", key, msgflg, res);
+        res
+  }
+}
+
+hook! {
+      unsafe fn msgctl(msqid: c_int, cmd: c_int, buf: *mut msqid_ds) -> c_int => msgctl_spy {
+        let res = real!(msgctl)(msqid, cmd, buf);
+        eprintln!("Spy: msgctl({:#x}, {:#x}, {:?}) => {}", msqid, cmd, buf, res);
+        res
+  }
+}
+
+hook! {
+    unsafe fn msgsnd(msqid: c_int, msgp: *const c_void, msgsz: size_t, msgflg: c_int) -> c_int => msgsnd_spy {
+        let res = real!(msgsnd)(msqid, msgp, msgsz, msgflg);
+        eprintln!("Spy: msgsnd({:#x}, {:?}, {}, {:#x}) => {}", msqid, msgp, msgsz, msgflg, res);
+        if msgsz == std::mem::size_of::<swtfb_ipc::swtfb_update>() {
+            let msg = &*(msgp as *const swtfb_ipc::swtfb_update);
+                eprintln!("Spy: msgsnd: Message: swt_update.mtype: {:?}, data: ... }}", msg.mtype);
+                let data_str_formatted = match msg.mtype {
+                    swtfb_ipc::MSG_TYPE::INIT_t => {
+                        format!("...")
+                    },
+                    swtfb_ipc::MSG_TYPE::UPDATE_t => {
+                        format!("{:?}", msg.data.update)
+                    },
+                    swtfb_ipc::MSG_TYPE::XO_t => {
+                        format!("{:?}", msg.data.xochitl_update)
+                    },
+                    swtfb_ipc::MSG_TYPE::WAIT_t => {
+                        format!("{:?}", msg.data.wait_update)
+                    }
+                };
+                eprintln!("Spy: msgsnd: Message: swt_update.data: {} }}", data_str_formatted);
+        }else {
+            eprintln!("Spy: msgsnd: Error: Message is not sizeof(swtfb_update) (expected {}, got {})", std::mem::size_of::<swtfb_ipc::swtfb_update>(), msgsz)
+        }
+        res
+  }
+}
--- a/examples/swtfb_sysv_spy.rs
+++ b/examples/swtfb_sysv_spy.rs
@@ -40,7 +40,7 @@ hook! {
                         format!("{:?}", msg.data.wait_update)
                     }
                 };
-                eprintln!("Spy: msgsnd: Message: swt_update.data: {} }}", data_str_formatted);
+                eprintln!("Spy: msgsnd: Message: swt_update.data: {}", data_str_formatted);
         }else {
             eprintln!("Spy: msgsnd: Error: Message is not sizeof(swtfb_update) (expected {}, got {})", std::mem::size_of::<swtfb_client::swtfb_update>(), msgsz)
         }
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -46,14 +46,12 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
                 .expect("Failed to open swtfb shared buffer");
             (device, Some(mem_map))
         } else {
-            (
-                OpenOptions::new()
-                    .read(true)
-                    .write(true)
-                    .open(path_to_device)
-                    .unwrap(),
-                None,
-            )
+            let device = OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(path_to_device)
+                .unwrap();
+            (device, None)
         };
 
         let mut var_screen_info = Framebuffer::get_var_screeninfo(&device, swtfb_client.as_ref());
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -12,7 +12,7 @@ use crate::framebuffer::common::{
     MXCFB_ENABLE_EPDC_ACCESS, MXCFB_SET_AUTO_UPDATE_MODE, MXCFB_SET_UPDATE_SCHEME,
 };
 use crate::framebuffer::screeninfo::{FixScreeninfo, VarScreeninfo};
-use crate::framebuffer::swtfb_ipc::SwtfbIpcQueue;
+use crate::framebuffer::swtfb_client::SwtfbClient;
 
 /// Framebuffer struct containing the state (latest update marker etc.)
 /// along with the var/fix screeninfo structs.
@@ -26,7 +26,7 @@ pub struct Framebuffer<'a> {
     /// like it has been done in `Framebuffer::new(..)`.
     pub var_screen_info: VarScreeninfo,
     pub fix_screen_info: FixScreeninfo,
-    pub swtfb_ipc_queue: Option<super::swtfb_ipc::SwtfbIpcQueue>,
+    pub swtfb_client: Option<super::swtfb_client::SwtfbClient>,
 }
 
 unsafe impl<'a> Send for Framebuffer<'a> {}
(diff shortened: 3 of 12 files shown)


src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+        screeninfo.blue.offset = 0;
+        screeninfo.blue.length = 5;
+        screeninfo
+    }
+}
+
+impl Drop for SwtfbClient {
+    fn drop(&mut self) {
+        if unsafe { libc::msgctl(self.msqid, libc::IPC_RMID, ptr::null_mut()) } != 0 {
+            panic!("Got an error when closing ipc queue!")
```
bkirwi, line 221: [IIUC](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics) panicking in `drop` will normally abort the process. Wondering if we should log here instead? IIUC this isn't a safety issue, so it might be more graceful to let the rest of shutdown proceed normally...

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+use std::io::Error as IoError;
+use std::os::unix::prelude::AsRawFd;
+use std::{env, mem, ptr};
+
+const SWTFB_MESSAGE_QUEUE_ID: i32 = 0x2257c;
+
+pub const WIDTH: i32 = 1404;
+pub const HEIGHT: i32 = 1872;
+
+#[allow(non_upper_case_globals)]
```
bkirwi, line 21: Why this instead of renaming or inlining?

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+#[allow(non_camel_case_types, dead_code)]
+pub struct wait_sem_data {
+    /// C string
+    pub sem_name: [u8; 512],
+}
+
+/// MSG_TYPE has to match swtfb_update_data !!!
+#[derive(Clone, Copy)]
+#[allow(non_camel_case_types, dead_code)]
+pub struct swtfb_update {
```
bkirwi, line 62: Should these be `#[repr(C)]`?

bkirwi (commented):
This is very cool!

I'm not familiar enough with the SysV stuff for a proper review, but I read through / left a few questions mostly for my own education. Feel free to ignore them if I'm missing the point. :)

I'll give this a spin on my own RM2, both with the standard demo app and a couple others, and report back. Hopefully within the next couple days.

src/framebuffer/swtfb_client.rs
LinusCDE: Good point! I didn't even know about this. A warn! is probably best here.

src/framebuffer/swtfb_client.rs
LinusCDE: The globals are copied/ported from the [source implementation](https://github.com/ddvk/remarkable2-framebuffer/) (mainly `ipc.cpp` and `client/main.cpp`). So I opted to keep most things as close a possible rather than making the names more rust like.

Since many things in the lib also use C-like naming, I think it fits the lib more anyway.

LinusCDE: Tbh those could probably be cleaned up though, since the global is duplicated anyway. This is probably a rather extreme case of keeping the names. It's literally the same variable in two naming schemes right after another. Didn't even notice that. Good catch!

src/framebuffer/swtfb_client.rs
LinusCDE: Wow! I somehow thought I thought of this, but It seems I never added it (or removed it during debugging?!?). Great catch. Making it explicit is always a lot better. Didn't even know that structs in rust have the memory layout as in C by default (maybe by chance)?

src/framebuffer/swtfb_client.rs
LinusCDE: Should be fixed with 9343082.

src/framebuffer/swtfb_client.rs
LinusCDE: Should be fixed with 88c97d7.

src/framebuffer/swtfb_client.rs
LinusCDE: Should be fixed with a0c0ea9.

LinusCDE:
> This is very cool!
> 
> I'm not familiar enough with the SysV stuff for a proper review, but I read through / left a few questions mostly for my own education. Feel free to ignore them if I'm missing the point. :)
> 
> I'll give this a spin on my own RM2, both with the standard demo app and a couple others, and report back. Hopefully within the next couple days.

Thanks for the review! I totally missed those parts. Nothing anything of this was related to how SysV is supposed to work. Mostly C and Rust tidbids, but important oversights and bad practice I didn't even know about. Huge thanks!

LinusCDE:
Made send_xochitl_update non public for now. Seems to not be meant for normal clients. I personally would still like the full structure for completness and e.g. `swtfb_sysv_spy.rs`.

Any idea how to proceed best?

https://discordapp.com/channels/385916768696139794/386181213699702786/909142226381176842

src/framebuffer/core.rs
```diff
@@ -25,20 +26,37 @@ pub struct Framebuffer<'a> {
...
 unsafe impl<'a> Sync for Framebuffer<'a> {}
 
 impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
     fn from_path(path_to_device: &str) -> Framebuffer<'_> {
-        let device = OpenOptions::new()
-            .read(true)
-            .write(true)
-            .open(path_to_device)
-            .unwrap();
+        let swtfb_client = if path_to_device == "/dev/shm/swtfb.01" {
```
fenollp, line 37: Let's define the string once and use it everywhere. Maybe as a method on the model gen enum?
```rust
impl Model {
    pub framebuffer_path(&self) -> &'static str {
        match *self {
            Model::Gen1 => "/dev/fb0",
            Model::Gen2 => "/dev/shm/swtfb.01",
        }
    }
}
```
thus the device `get_framebuffer_path()` also becomes:
```rust
    pub fn get_framebuffer_path(&self) -> &'static str {
        self.model.framebuffer_path()
    }
```
and here: (this is surely wrong syntax but you get the idea)
```rust
        let swtfb_client = if path_to_device == (Model::Gen2).framebuffer_path() {
```

src/framebuffer/core.rs
```diff
@@ -25,20 +26,37 @@ pub struct Framebuffer<'a> {
...
+            (device, Some(mem_map))
+        } else {
+            (
+                OpenOptions::new()
+                    .read(true)
+                    .write(true)
+                    .open(path_to_device)
+                    .unwrap(),
+                None,
+            )
```
fenollp, line 56:
```suggestion
            let device = OpenOptions::new()
                .read(true)
                .write(true)
                .open(path_to_device)
                .unwrap();
            (device, None)
```

src/framebuffer/core.rs
```diff
@@ -75,10 +97,16 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
             default_font,
             var_screen_info,
             fix_screen_info,
+            swtfb_client,
         }
     }
 
     fn set_epdc_access(&mut self, state: bool) {
+        if self.swtfb_client.is_some() {
+            // Not catched in rm2fb => noop
```
fenollp, line 106:
```suggestion
            // Not handled in rm2fb => noop
```

src/framebuffer/core.rs
```diff
@@ -103,6 +137,11 @@ impl<'a> framebuffer::FramebufferBase<'a> for Framebuffer<'a> {
     }
 
     fn set_update_scheme(&mut self, scheme: u32) {
+        if self.swtfb_client.is_some() {
+            // Not catched in rm2fb => noop
```
fenollp, line 141:
```suggestion
            // Not handled in rm2fb => noop
```

src/framebuffer/refresh.rs
```diff
@@ -126,34 +120,32 @@ impl<'a> framebuffer::FramebufferRefresh for core::Framebuffer<'a> {
...
-                markerdata.collision_test
+                self.wait_refresh_complete(whole.update_marker)
             }
             PartialRefreshMode::Async => whole.update_marker,
         }
     }
 
     fn wait_refresh_complete(&self, marker: u32) -> u32 {
+        if let Some(ref swtfb_client) = self.swtfb_client {
+            swtfb_client.wait_for_update_complete();
```
fenollp, line 144:
```suggestion
            return swtfb_client.wait_for_update_complete();
```
and move the comment there under a TODO (assuming this could be achieved in the future?)

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,227 @@
...
+use std::ffi::{c_void, CString};
+use std::fs::{File, OpenOptions};
+use std::io::Error as IoError;
+use std::os::unix::prelude::AsRawFd;
+use std::{env, mem, ptr};
+
+const SWTFB_MESSAGE_QUEUE_ID: i32 = 0x2257c;
+
+pub const WIDTH: i32 = 1404;
+pub const HEIGHT: i32 = 1872;
```
fenollp, line 20: Why not reuse `common::DISPLAYHEIGHT`?

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,227 @@
...
+use memmap2::{MmapOptions, MmapRaw};
+use std::ffi::{c_void, CString};
+use std::fs::{File, OpenOptions};
+use std::io::Error as IoError;
+use std::os::unix::prelude::AsRawFd;
+use std::{env, mem, ptr};
+
+const SWTFB_MESSAGE_QUEUE_ID: i32 = 0x2257c;
+
+pub const WIDTH: i32 = 1404;
```
fenollp, line 19: Why not reuse `common::DISPLAYWIDTH`?

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,227 @@
...
+pub const HEIGHT: i32 = 1872;
+
+pub const BUF_SIZE: i32 = WIDTH * HEIGHT * std::mem::size_of::<u16>() as i32; // hardcoded size of display mem for rM2
+const SEM_WAIT_TIMEOUT: i32 = 200000000; /* 200 * 1000 * 1000, e.g. 200ms */
```
fenollp, line 23:
```suggestion
const SEM_WAIT_TIMEOUT_NS: i32 = 200_000_000;
```

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+        };
+        assert!(msqid >= 0);
+
+        // Does nested need special handling?
+        // This may not be needed at all.
+        if env::var("RM2FB_ACTIVE").is_ok() {
+            env::set_var("RM2FB_NESTED", "1");
+        } else {
+            env::set_var("RM2FB_ACTIVE", "1");
+        }
```
fenollp, line 98: No idea if it's needed. How does one can produce the nested case?
```cpp
// from ddvk's ipc.cpp : nesting only triggers logging a line
  uint16_t *mem =
      (uint16_t *)mmap(NULL, BUF_SIZE, PROT_WRITE, MAP_SHARED, fd, 0);

  if (getenv("RM2FB_NESTED") == "") {
    fprintf(stderr, "OPENED SHARED MEM: /dev/shm%s at %x, errno: %i\n",
            name.c_str(), mem, errno);
  }
  return mem;
```

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+impl SwtfbClient {
+    pub fn open_buffer(&self) -> Result<(File, MmapRaw), IoError> {
+        let device = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/dev/shm/swtfb.01")?;
```
fenollp, line 112: Ditto on magic strings / DRY.

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub fn wait_for_update_complete(&self) {
+        if !self.do_wait_ioctl {
+            return;
+        }
+
+        // https://github.com/ddvk/remarkable2-framebuffer/blob/1e288aa9/src/client/main.cpp#L149
+
+        let sem_name_str = format!("/rm2fb.wait.{}", unsafe { libc::getpid() });
+        let mut sem_name = [0u8; 512];
```
fenollp, line 154:
```suggestion
        let mut sem_name = [0u8; 12 + 10];
```
`pid_t` is `i32` => 10 digits tops

src/framebuffer/swtfb_client.rs
```diff
@@ -0,0 +1,224 @@
...
+        };
+        unsafe {
+            libc::clock_gettime(libc::CLOCK_REALTIME, &mut timeout);
+        }
+        timeout.tv_nsec += SEM_WAIT_TIMEOUT;
+
+        if timeout.tv_nsec >= 1_000_000_000 {
+            timeout.tv_nsec -= 1_000_000_000;
+            timeout.tv_sec += 1;
+        }
```
fenollp, line 174: Shouldn't this be in a loop?
I have no idea if the clock here can jump around forward/backwards in large/small leaps...

fenollp (commented):
Great work! I should be able to test this on both archs soon, I'm excited.
I think you're right about using `thiserror` and also about refactoring to implement swfb as traits. 

Not sure why there was discussion on dynamic dispatch for traits though? but sure do whatever

examples/swtfb_sysv_spy.rs
```diff
@@ -0,0 +1,49 @@
...
+                    },
+                    swtfb_client::MSG_TYPE::XO_t => {
+                        format!("{:?}", msg.data.xochitl_update)
+                    },
+                    swtfb_client::MSG_TYPE::WAIT_t => {
+                        format!("{:?}", msg.data.wait_update)
+                    }
+                };
+                eprintln!("Spy: msgsnd: Message: swt_update.data: {} }}", data_str_formatted);
+        }else {
```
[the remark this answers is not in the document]
LinusCDE: Right. Stuff in a macro can be anything, so I think code highlighting is harder for linters.

Had a stray "}}" there. But should still work nicely though.

src/framebuffer/core.rs
LinusCDE: Good point!

examples/swtfb_sysv_spy.rs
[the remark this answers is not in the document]
LinusCDE: Should be done with commit 3dc5f92.

src/framebuffer/core.rs
LinusCDE: Should be done with commit 6f270a8.

src/framebuffer/core.rs
LinusCDE: I believe that "catched" describes it better here. Since the ioctl wrapper didn't return its own result, the ioctl is passed on to whereever [here](https://github.com/ddvk/remarkable2-framebuffer/blob/1e288aa9cae24f7dda469bfaa32d92def90ba0e9/src/client/main.cpp#L232).

So it's not only just not handles, but the ioctl could technically still go through to the device/file if it were to be supported. Maybe stating something like "catched/handled" instead?

src/framebuffer/refresh.rs
LinusCDE: This would cause a syntax error as `swtfb_client.wait_for_update_complete()` returns nothing. There is a lot of mostly copied c code that in the original client also didn't assume any error condition. Probably adding such to the function would be better first.

fenollp:
WRT `send_xochitl_update`: if it helps with reverse engineering I'm fine with it.
 It's out of the spec so let's just document that and it's all good :)

src/framebuffer/swtfb_client.rs
LinusCDE: Since that could become conditional to a model if a potential Gen 3 changes the resolution (as already happened to MT and Wacom sizes). While this code is duplicated, this should prevent potential problems should such a change occur. And if they continue using a sw based fb, the client would need an update/newer client anyway.

From a technical point, DISPLAYWIDTH should already be a conditional value (as I [learned from reStream](https://github.com/rien/reStream/blob/9deee838/reStream.sh#L131)). libremarkable just always "hid" those last 4 pixels and got away with it.

In that case this would also break, since a const can't be derived from a static once_cell. We actually could address this in a future release.

src/framebuffer/swtfb_client.rs
LinusCDE: I hadn't bothered redoing this with rust types yet. Thought of this, too but rm2fb [did the same](https://github.com/ddvk/remarkable2-framebuffer/blob/1e288aa9/src/client/main.cpp#L175) and got away for it. For the sake of not needing to test this myself, the code is basically a copy for now.

I also had done some things like that for work. There are weird cases where such stuff is meant for leap seconds and rust's chrono and/or time crates handle conversion from timespec a lot more strictly (even though this just seems like a tame overflow counter). So didn't really wanna tamper with the already working solution. But your point seems pretty important as I know think about it more.

@raisjn any insights here? May be an issue for your client as well. Maybe this never mattered since the timeout never exceeded it in testing?

LinusCDE:
> WRT `send_xochitl_update`: if it helps with reverse engineering I'm fine with it. It's out of the spec so let's just document that and it's all good :)

Agreed. I'll probably still like to keep the function private (or just remove it). You can just construct a `swtfb_update` easily yourself and send with the public `send()` method anyway. So the lib wouldn't actually prevent you from doing this, just require a bit more knowledge which seems like a fine trade-off.

src/framebuffer/core.rs
fenollp: Well it's just that past tense of "catch" is "caught". Do whatever :)

src/framebuffer/refresh.rs
fenollp: ofc what I'm saying is having `wait_for_update_complete` return the error code so 
```rust
            // Assume success
            return 0;
```
is within `wait_for_update_complete`'s body i.e. geographically close to the logic.

src/framebuffer/swtfb_client.rs
fenollp: I'd rather we don't repeat magic numbers when we have `crate::framebuffer::common::DISPLAYHEIGHT` today.

src/framebuffer/core.rs
LinusCDE: Should be okay now with commit 79816f01.

src/framebuffer/refresh.rs
LinusCDE: While I would agree in general, I disagree for this case. Rm2fb doesn't support the return type which is a collision check. Adding it within swtfb_client would suggest a feature that is doesn't have. The method is also slightly different phrased to match the ioctl name that rm2fb handles, not libremarkables one.

While the ioctl technicalla gets the marker data and "returns" the collision value (u32), I would rather the missing support to be obvious here (rm2fb doesn't care about the marker).

src/framebuffer/swtfb_client.rs
LinusCDE: Good point. Should be done with commit a4e3fd58.

src/framebuffer/swtfb_client.rs
LinusCDE: For now I think we should just keep it as close to the original client as possible. If the client changes this, we can do so as well later. But it may be a bug that just got never triggered.

src/framebuffer/swtfb_client.rs
LinusCDE: Fixed in commit a281c97.

src/framebuffer/swtfb_client.rs
LinusCDE: Not sure either. Seems like a debugging case but the behaviour isn't strictly altered when a nested app is detected. The [spec/design](https://github.com/ddvk/remarkable2-framebuffer/issues/4) also doesn't mention anything here.

src/framebuffer/swtfb_client.rs
raisjn: I don't remember the reasoning here or why we did this, but a loop sounds right. there might be something in the git log for rm2fb about it (use blame?)

LinusCDE: The performance could probably be even faster/branchless when using something like this instead:

```diff
timeout.tv_nsec += SEM_WAIT_TIMEOUT;
+ timeout.tv_sec += timeout.tv_nsec / 1e9;
+ timeout.tv_nsec %= 1e9;
- if (timeout.tv_nsec >= 1e9) {
-   timeout.tv_nsec -= 1e9;
-   timeout.tv_sec++;
- }
```

Makefile
```diff
@@ -1,5 +1,5 @@
-# For musl, use: armv7-unknown-linux-musleabihf
-TARGET ?= armv7-unknown-linux-gnueabihf
+# For non-musl, use: armv7-unknown-linux-gnueabihf
+TARGET ?= armv7-unknown-linux-musleabihf
```
fenollp, line 2: I should probably revert this as this value is used in this Makefile for `cargo` calls, not just `cross` calls...

LinusCDE: Sounds solid. I'm fine with either way.

src/framebuffer/swtfb_client.rs
LinusCDE: Added the above change with commit 3bd976cf.

To get this done here I opened a PR here for the original client to address it separately there: https://github.com/ddvk/remarkable2-framebuffer/pull/81

src/framebuffer/swtfb_client.rs
LinusCDE: Is handling this important in the client @raisjn ? Seems to be merely used for more accurate debug output.

raisjn: it's not necessary, mostly was about debugging some stuff. not even sure nesting works correctly. i think rmkit uses RM2FB env variable to detect if it is on rm2, but don't think libremarkable needs it

LinusCDE: Okay. Remove this with commit 6a8dbe16 for now then. If this is ever needed for advanced checking of errors or such, we can add it again.

LinusCDE:
I think all standing points should be resolved now.

bkirwi:
Nice work!

I'm personally comfortable with merging... any API cleanup / etc. we decide we want can happen in future PRs.

PR #79, opened 2021-11-16
rM1+rm2fbc: fix code relating to logging refresh errors + musl CI

fenollp:


LinusCDE:
Works fine. :+1:

(Btw just feel free to commit directly into that branch. I put it in upstream (this repo) to explicitly invite this, since it's way faster.)

PR #81, opened 2021-12-10
Impl Copy, Clone for all types in common.rs

LinusCDE:
This is a minor fix that allows to Copy and Clone all types in `src/framebuffer/common.rs`. More a minor thing to improve usability for some cases.

I originally stumbled upon this when doing [doomarkable](https://github.com/LinusCDE/doomarkable). Wanted to specify the waveform depending on the device and later reuse it depending on the model in a render loop. Problem was that the variable in loop wouldn't work, since the parameter is owned and rusts ownership wouldn't permit me to give ownership repeatedly away in a loop (duh).

Wouldn't be a problem if the waveform mode could be cloned or such, but this wasn't derive. So I would basically need to create my own clone, remember it otherwise (basically serializing to a different type (e.g. and int) deserializing it again every time). My workaround was to just find out the same value every time [in the loop](https://github.com/LinusCDE/doomarkable/blob/9b5785ba4a52679dfb39900b78a019bfa30b47ef/src/main.rs#L251) while it would be perfectly fine to do once [before the loop](https://github.com/LinusCDE/doomarkable/blob/9b5785ba4a52679dfb39900b78a019bfa30b47ef/src/main.rs#L189) (any maybe use a clone on that value if not also Copy).

Clone would be enough, but those types are pretty primitive and surely not worth the cloning. Also more complex types in the same file already had also "Copy" specifies. So I took the liberty to ensure that all types in that file have Copy and Clone.

For the next version, looking that everything is at least Debug'able and (in most cases) at Clone'able would probably be good as well. Just wanted to to this quick fix now since that note sat on my todo list ever since.

Maybe those types should also get PartialEq, Eq and Hash as well in case people would need it. Not sure what Rust's philosophy would be in such a case (better to "over-derive" or "under-derive"?).

### DIFF
--- a/src/framebuffer/common.rs
+++ b/src/framebuffer/common.rs
@@ -300,20 +300,20 @@ pub enum mxcfb_ioctl {
     MXCFB_ENABLE_EPDC_ACCESS = 0x36,
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum auto_update_mode {
     AUTO_UPDATE_MODE_REGION_MODE = 0,
     AUTO_UPDATE_MODE_AUTOMATIC_MODE = 1,
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 pub enum update_scheme {
     UPDATE_SCHEME_SNAPSHOT = 0,
     UPDATE_SCHEME_QUEUE = 1,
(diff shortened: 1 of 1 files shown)


fenollp:
I don't have an opinion. I'm fine with the changes. Usually I'm in the team "let's add things as we need them" and not "we might need this in the future so let's add them now".

LinusCDE:
Fine with me as well. I think these changes should suffice for the foreseeable future.

PR #82, opened 2021-12-12
Add a screenshot utility

bkirwi:
A quick screenshot script; it writes the screen contents as a PNG to the given path, or to stdout if no path is given.

This is a small amount of code, but seemed generally useful enough -- and a simple enough usage of the library -- to be worth contributing. But if y'all disagree I don't feel super strongly about it.

### DIFF
--- a/examples/screenshot.rs
+++ b/examples/screenshot.rs
@@ -1,4 +1,5 @@
 use image::{DynamicImage, ImageOutputFormat};
+use libremarkable::device::CURRENT_DEVICE;
 use libremarkable::framebuffer::common::*;
 use libremarkable::framebuffer::core::*;
 use libremarkable::framebuffer::*;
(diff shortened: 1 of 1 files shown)


examples/screenshot.rs
```diff
@@ -0,0 +1,48 @@
+use image::{DynamicImage, ImageOutputFormat};
+use libremarkable::framebuffer::common::*;
+use libremarkable::framebuffer::core::*;
+use libremarkable::framebuffer::*;
+use libremarkable::image::RgbImage;
+use std::fs::OpenOptions;
+
+fn main() {
+    let fb = Framebuffer::from_path("/dev/fb0");
```
LinusCDE, line 9:
```suggestion
    let fb = Framebuffer::from_path(crate::device::CURRENT_DEVICE.get_framebuffer_path());
```

Autoselect the framebuffer so `rm2fb-client` is not needed when executing the command.

LinusCDE (approved):
This seems like a really useful addition. I'm for it.

examples/screenshot.rs
bkirwi: Good call - done.

PR #83, opened 2021-12-16
Bump Rust edition, drop LTO and update deps: once_cell, atomic

fenollp:


bkirwi (approved):
Good call!

fenollp:
Note: I tested this on both rM1 and rM2.

PR #87, opened 2022-01-06
New event loop usage

LoganDark:
The old model was hard to use and not very Rusty, so I decided to make it work a bit more like winit.

Also comes with a new demo showcasing how easy it is to use the new event loop to build a simple program.

### DIFF
--- a/examples/basic_draw.rs
+++ b/examples/basic_draw.rs
@@ -1,0 +1,84 @@
+//! This example is a very basic drawing application. Marker can draw on the
+//! screen (without tilt or pressure sensitivity) and Marker Plus can use the
+//! eraser end to erase.
+//!
+//! Drawing is done in the framebuffer without any caching, so it's not possible
+//! to save the results to file, zoom or pan, etc. There are also no GUI
+//! elements or interactivity other than the pen.
+//!
+//! The new event loop design makes this type of application very easy to make.
+
+use libremarkable::appctx::ApplicationContext;
+use libremarkable::framebuffer::{FramebufferDraw, FramebufferRefresh};
+use libremarkable::framebuffer::common::{color, display_temp, dither_mode, DRAWING_QUANT_BIT, waveform_mode};
+use libremarkable::framebuffer::refresh::PartialRefreshMode;
+use libremarkable::input::InputEvent;
+use libremarkable::input::wacom::{WacomEvent, WacomPen};
+
+fn main() {
+	let mut app = ApplicationContext::new();
+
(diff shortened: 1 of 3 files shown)


LinusCDE (commented):
The changes seem reasonable and the example works fine.

Some points:
 - Add some text, telling you to draw would be cool. Got confused at first since nothing really happened (already had a blank screen though)
 - Address clippy and rustfmt failing.

I'm personally fine with the way this works. @bkirwi, @fenollp any thoughts here?

Screenshot:

LoganDark:
I've checked "allow edits by maintainers", feel free to make any changes you see fit. I think the rustfmt fail is because I use tabs instead of spaces - I'd rather leave the conversion to someone else, since I don't like spaces.

LinusCDE:
Since there don't seem any additional comments from other maintainers, I'll just consider this change fine and merge it.

bkirwi:
I personally don't have strong feelings about the change. I don't love the callbacks in `appctx`, and this looks like it might make it easier to ditch some of the awkward `transmute`ing down the road, but the new demo app seems to be about as easy to write if you avoided the `appctx` stuff entirely and just used the raw inputs and the framebuffer stuff directly. (Which IIUC is how most people are using the library anyways, including me!)

It's a bit awkward that libremarkable is a mix of low-level primitives for writing on the remarkable, plus some opinionated higher level stuff that many apps end up outgrowing. There may be more ways to tweak `appctx` such that it works well for more apps... but maybe it's worth fleshing out the documentation to make it more clear what parts are essential and which parts are easily ignored?

LoganDark:
> seems to be about as easy to write if you avoided the appctx stuff entirely and just used the raw inputs and the framebuffer stuff directly.

That's exactly what is happening except it's more difficult to set that stuff up without the appctx. appctx gives you events. Without it you have to start threads and stuff manually, so it's easier to just use appctx. You don't have to use anything except the event loop.

bkirwi:
Sure. For posterity, it's something like:

```rust
let (input_tx, input_rx) = channel::<InputEvent>();
EvDevContext::new(InputDevice::GPIO, input_tx.clone()).start();
EvDevContext::new(InputDevice::Multitouch, input_tx.clone()).start();
EvDevContext::new(InputDevice::Wacom, input_tx).start();
while let Ok(event) = input_rx.recv() {
    // ~whatever
}
```

(And drop some of the input devices if you don't care about those events, of course.)

`appctx` saves you a couple lines! But it's fairly close.

PR #90, opened 2022-01-13
make Roboto optional

LoganDark:
This means you won't pay the cost of

- font storage in the binary
- up-front font parsing at runtime
- storage of the parsed structures at runtime

if you don't use the built-in font drawing code.


LinusCDE (approved):
Demo runs fine

PR #95, opened 2022-01-24
Make rgb565 conversion algorithms a bit better

LoganDark:
The old algorithm has some unpleasant tinting. Native components `[0xFF, 0xFF]` don't result in rgb components `[0xFF, 0xFF, 0xFF]` but rather `[0xF8, 0xFC, 0xF8]`. This affects screenshots for example.

Obviously some fudging is needed in order to come up with the "extra resolution" in rgb8 to make the mins and maxes match up. Roundtripping from 565 -> 888 -> 565 is still a thing. And now it's a part of the test suite, which verifies that every rgb565 value can be roundtripped to rgb8 and back. And that rgb565 -> rgb8 conversions behave properly.

The new conversion routines still only use integer arithmetic. I haven't benchmarked them but they should actually be slightly faster than the old routines.

Note that I tested these on my desktop machine which required #92 in order to get libremarkable to build. Just a preview of how useful that can be.

### DIFF
--- a/src/framebuffer/common.rs
+++ b/src/framebuffer/common.rs
@@ -110,12 +110,12 @@ impl color {
 #[test]
 fn rgb565_conversions() {
     // Ensure that min and max values are transformed faithfully
-    assert_eq!(color::RGB(0, 0, 0).to_rgb565(), [0x0, 0x0]);
+    assert_eq!(color::RGB(0, 0, 0).to_rgb565(), [0, 0]);
     assert_eq!(color::RGB(255, 255, 255).to_rgb565(), [255, 255]);
     assert_eq!(color::from_native([0, 0]).to_rgb8(), [0, 0, 0]);
     assert_eq!(color::from_native([255, 255]).to_rgb8(), [255, 255, 255]);
     assert_eq!(color::GRAY(0).to_rgb565(), [255, 255]);
-    assert_eq!(color::GRAY(255).to_rgb565(), [0x0, 0x0]);
+    assert_eq!(color::GRAY(255).to_rgb565(), [0, 0]);
 
     // Ensure that every single RGB565 value can be transformed to RGB8 and back losslessly
     for native in 0..u16::MAX {
--- a/src/framebuffer/common.rs
+++ b/src/framebuffer/common.rs
@@ -60,11 +60,11 @@ impl color {
     }
 
     pub fn to_rgb8(self) -> [u8; 3] {
-        let rgb565 = u16::from_be_bytes(self.as_native());
+        let rgb565 = u16::from_le_bytes(self.as_native());
 
-        let r5 = rgb565 & 0b11111;
+        let r5 = rgb565 >> 11 & 0b11111;
         let g6 = rgb565 >> 5 & 0b111111;
-        let b5 = rgb565 >> 11 & 0b11111;
+        let b5 = rgb565 & 0b11111;
 
         let r8 = (r5 * 255 / 0b11111) as u8;
         let g8 = (g6 * 255 / 0b111111) as u8;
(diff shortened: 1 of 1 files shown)


LoganDark:
Maybe screw you rustfmt and your completely arbitrary rules. The function looks fine the way it is and applying that patch would make it worse

LoganDark:
This actually fixes a bug in `color::to_rgb8` that results in r and b being swapped

src/framebuffer/common.rs
```diff
@@ -97,14 +97,32 @@ impl color {
...
+
+#[test]
+fn rgb565_conversions() {
+    // Ensure that min and max values are transformed faithfully
+    assert_eq!(color::RGB(0, 0, 0).to_rgb565(), [0, 0]);
+    assert_eq!(color::RGB(255, 255, 255).to_rgb565(), [255, 255]);
+    assert_eq!(color::from_native([0, 0]).to_rgb8(), [0, 0, 0]);
+    assert_eq!(color::from_native([255, 255]).to_rgb8(), [255, 255, 255]);
+    assert_eq!(color::GRAY(0).to_rgb565(), [255, 255]);
+    assert_eq!(color::GRAY(255).to_rgb565(), [0, 0]);
```
bkirwi, line 118: We should eventually add some test here that ensures we don't get the color channels swapped... catch any future bugs like the ones you fixed in the last couple commits. Up to you whether you want to do that before merging or not - this is obviously a lot better than what we had before.

bkirwi (approved):
Solid improvement, and thanks for the tests!

Haven't heard of any issues building on desktop before... which dep failed to build / any idea why?

src/framebuffer/common.rs
LoganDark: I can hardcode RGB values for a couple of the built-in colors like RED, if that works? :)

LoganDark:
I had issues building evdev and epoll since I was cross-compiling for Windows (WSL 1 - I don't like VMs - no graphics support). #92 adds the ability to exclude those from the build. Although it looks like nobody's wanted to touch it for a while - now that I depend on it, I need it merged at some point but I fear that it's too big/complicated now.

I've been testing this by merging it into #92 on my local checkout and then running `cargo test --target x86_64-unknown-linux-gnu --package libremarkable --lib framebuffer::common::rgb565_conversions --no-default-feature
s --features framebuffer-types -- --exact`. It's a mouthful but it works!

src/framebuffer/common.rs
bkirwi: Works for me!

LoganDark: I'll also throw in an `assert_eq!(color::RGB(255, 0, 0).to_rgb8(), [255, 0, 0])` for good measure

bkirwi:
> I had issues building evdev and epoll since I was cross-compiling for Windows (WSL 1 - I don't like VMs - no graphics support).

Make sense - thanks!

src/framebuffer/common.rs
LoganDark: Turns out I still need to figure out some endianness stuff after all. As long as `color::RED` is accurate and those native components actually correspond to red... I'll figure it out! lol

LoganDark:
I added tests for the specific colors, and I'm 100% sure that red/green were swapped. I'm also 100% sure that the endianness is CORRECT this time - both in terms of the byte order of the u16 and the order of the components in that u16. Test passes. Phew........

LoganDark:
You should probably squash all these commits when you merge.

src/framebuffer/common.rs
LoganDark: Tests have been added, algos corrected for the third (fourth?) time

LoganDark:
By the way: Are we sure that RED and BLUE aren't swapped? Because those are the basis of the r,g,b orderings and it's important to get this right. The bit patterns of RED and BLUE are exact reverses of each other so it's not possible to deduce whether they're correct just from that.

LoganDark:
Just for the record the endianness is now correct and proven to work correctly on-device. Since the encoding and decoding algorithms match, I was able to run `screenshot` with #96 to get this beauty:

image

to prove beyond a shadow of a doubt that YES, we understand the rgb565_le format correctly. (`[gggbbbbb, rrrrrggg]`)

For those following along at home, xochitl only writes color information to the framebuffer if screenshare is turned on.

fenollp (approved):

fenollp:
@bkirwi Next time maybe select the "Squash and merge" option?

PR #94, opened 2022-01-17
Add more image formats to "live" example and a html selection page

LinusCDE:
I usually used a "livepng" version where I replaced the "JpegEncoder" with an "PngEncoder" since the images where usually smaller and lossless for my usecases (usually screenshots for READMEs in my projects).

Started this out with the intend to add png support alongside jpeg. Since we're using a webserver, I also decided to add a minimal html page to allow users to choose what format to use. Also checked that it renders okay on smartphones.

You can also now see the duration of the encoding in the console.
grafik

PR #96, opened 2022-01-25
Perform gamma correction in screenshot.rs

LoganDark:
Right now, screenshot.rs takes the raw rgb values and pipes them into a PNG image.

This is not sufficient; viewers generally assume that pixel data in PNGs is sRGB (unless otherwise specified, which `screenshot.rs` does not). So map it from linear RGB to sRGB. This results in graphics looking much smoother on computer monitors.


LoganDark:
Non-gamma-corrected (current behavior, left) and gamma-corrected (PR behavor, right) screenshots, side by side:



Full size (if you have a 1x screen):



The one on the right looks closer to how the reMarkable physically displays the framebuffer. It's just that the PNG image had to be prepared (gamma correction) to properly look that way in an image viewer, which generally expects sRGB from PNG images. And notice how the outlines around the buttons look a lot more consistent on the gamma-corrected screenshot.

fenollp (approved):

LoganDark:
Not sure how you'll feel about the last commit but I thought it'd be worth pushing so you can try it out. I'll probably need to revert it before this is merged :P

LoganDark:
It might be worth incorporating a couple LUTs like that one into libremarkable itself to provide convenience functions like, for example, `to_srgb`, or `from_srgb`.

examples/screenshot.rs
```diff
@@ -6,28 +6,8 @@ use libremarkable::framebuffer::*;
...
-        221, 222, 222, 223, 223, 224, 224, 225, 225, 226, 226, 227, 227, 228, 228, 229, 229, 230,
-        230, 231, 231, 232, 232, 233, 233, 234, 234, 235, 235, 236, 236, 237, 237, 238, 238, 239,
-        239, 240, 240, 241, 241, 242, 242, 243, 243, 244, 244, 245, 245, 246, 246, 247, 247, 248,
-        248, 249, 249, 249, 250, 250, 251, 251, 252, 252, 253, 253, 254, 254, 255,
-    ];
-
-    [GAMMA[r as usize], GAMMA[g as usize], GAMMA[b as usize]]
-}
+mod rgb565_to_srgb;
+use rgb565_to_srgb::RGB565_TO_RGB888;
```
fenollp, line 10: At this point wouldn't it make sense to actually publish this where it can be reused by other things than this example?

LoganDark: Yes, it would, which is actually why I've been working on a crate for it for the past few hours. It should be able to replace libremarkable's home-grown rgb565 functions and also provide some runtime efficiency via LUTs generated at compile-time.

It includes rgb565_le, as well as some common transformations with optional embedded look-up tables.

fenollp: Cool.
Dang I'm curious what the compile times are on your https://docs.rs/rgb565/latest/rgb565/

LoganDark: A cold debug build takes 1.32s. A cold release build takes 2.24s. Without the 32MB LUTs enabled of course. With those enabled, debug build takes 21s and release build takes 17.06s. But they're disabled by default.

LoganDark: A cold debug build takes 1.32s. A cold release build takes 2.24s. Without the 32MB LUTs enabled of course. With those enabled, debug build takes 21s and release build takes 17.06s. But they're disabled by default.

LoganDark:
Embedded LUT:

```
real    0m0.560s
user    0m0.506s
sys     0m0.050s
```

`rgb565` crate:

```
real    0m0.578s
user    0m0.556s
sys     0m0.021s
```

Not sure where the mismatch is, but it's close enough. Still faster than the original (0.679s)!

LoganDark:
Is this good to merge? Anything more I need to address?

LinusCDE (approved):
Tested it out and it works fine!

I should also consider adding this to #94 then.

PR #99, opened 2022-01-31
msgsz should measure only the message contents

### LINKED ISSUE #98: Buffer overread in swtfb client
There is a buffer overread in the following call to `msgsnd`:
https://github.com/canselcik/libremarkable/blob/b146c390af05f5e69529e0ee5c635767e2e8b1e8/src/framebuffer/swtfb_client.rs#L125

As per [the man page](https://man7.org/linux/man-pages/man2/msgop.2.html), the `msgsz` argument (third argument) should not include the size of the mtype field, only of the data field, so currently this call will read 4 extra bytes from memory and send them to the server.

(Related to ddvk/remarkable2-framebuffer#89.)

bkirwi:
Fixes #98.

@matteodelabre - does this look like the proper fix, or is there more to do?

### DIFF
--- a/src/framebuffer/swtfb_client.rs
+++ b/src/framebuffer/swtfb_client.rs
@@ -122,7 +122,7 @@ impl SwtfbClient {
     pub fn send(&self, update: &swtfb_update) -> bool {
         unsafe {
             let ptr = ptr::addr_of!(*update) as *const c_void;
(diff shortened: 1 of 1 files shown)


matteodelabre:
Yes, this looks good!

bkirwi:
Thanks!

Issue #101, opened 2022-04-08
Linking Error


ju6ge:
Hey I have been trying to build the code myself, because I am interested in playing around and maybe creating an application for the remarkable. I have installed both SDKs ("codex", "oecore"). But no matter which one i set in the cargo config the build of the examples fails with a linker error. Not really sure what to make of it. Any ideas what I am missing?

```
/home/judge/GitHub/remarkable/toolchains/oecore-x86_64/sysroots/x86_64-oesdk-linux/usr/libexec/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/7.3.0/real-ld: cannot find Scrt1.o: No such file or directory
          /home/judge/GitHub/remarkable/toolchains/oecore-x86_64/sysroots/x86_64-oesdk-linux/usr/libexec/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/7.3.0/real-ld: cannot find crti.o: No such file or directory
          /home/judge/GitHub/remarkable/toolchains/oecore-x86_64/sysroots/x86_64-oesdk-linux/usr/libexec/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/7.3.0/real-ld: cannot find crtbeginS.o: No such file or directory
          collect2: error: ld returned 1 exit status
```

LinusCDE:
I have not seen this particular linking error before.

Some ideas:

 - Ensure the `environment-setup-cortexa9hf-neon-oe-linux-gnueabi` file in your toolchain location is sourced
 - Disable the appctx feature (since it links lua stuff it's the most likely to cause linker problems if cross-compilation is not set up properly): `libremarkable = { version = "0.6.0", default-features = false, features = [ "framebuffer" ] )
 - Investigate more by adding `-vv` for more debugging info
 - Try to just use cross

ju6ge:
So I made sure to start from a clean environment and source the correct environment file. But the linker error persists. The files are definitely there. They are located under
`<sysroot>/usr/lib`. Just the linker refuses to find them. I tried setting an linker option for the sysroot as well, which did not help …

Disabling appctx is not an option since for the start I am just trying to build the demo app. I will need to investigate further and maybe also try to use a different machine. But anyway I have some more pressing things on my plate right now 🙈.

LinusCDE:
Just to narrow down the problem, it would be helpful to disable appctx nonetheless.

Let me know when you investigate further on this.

[...]

PR #102, opened 2022-04-11
added script to generate .cargo/config from toolchain environment 💪

ju6ge:
This commit adds a python script to generate the correct .cargo/config for the installed toolchain. All path are extracted from the environment variables after sourcing the toolchain's environment file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ In order to build `libremarkable` and the examples (`spy.so` and `demo`), you'll
 
 You can then set up your Rust toolchain for cross compilation with: `rustup target add armv7-unknown-linux-gnueabihf`.
 
-In order for rust to leverage the toolchain a `.cargo/config` file is required. This file can be generated using the `gen_cargo_config.py`. First the toolchain environment must be
-sourced. It location is can be found within the toolchain installation directory. The correct path is also referenced in the toolchain [wiki](https://remarkablewiki.com/devel/toolchain).
+In order for rust to leverage the toolchain a `.cargo/config` file is required. This file can be generated using `gen_cargo_config.py`. First the toolchain environment must be
+sourced. Its location is can be found within the toolchain installation directory. The correct path is also referenced in the toolchain [wiki](https://remarkablewiki.com/devel/toolchain).
 After the environment is loaded the script will read the environment variables to generate the correct `.cargo/config` file for your toolchain.
 
 The resulting config file will look something like this:
--- a/README.md
+++ b/README.md
@@ -27,6 +27,20 @@ In order for rust to leverage the toolchain a `.cargo/config` file is required. 
 sourced. It location is can be found within the toolchain installation directory. The correct path is also referenced in the toolchain [wiki](https://remarkablewiki.com/devel/toolchain).
 After the environment is loaded the script will read the environment variables to generate the correct `.cargo/config` file for your toolchain.
 
+The resulting config file will look something like this:
+```
+[target.armv7-unknown-linux-gnueabihf]
+linker = "<toolchain_install_path>/sysroots/x86_64-codexsdk-linux/usr/bin/arm-remarkable-linux-gnueabi/arm-remarkable-linux-gnueabi-gcc"
+rustflags = [
+  "-C", "link-arg=-march=armv7-a",
+  "-C", "link-arg=-marm",
+  "-C", "link-arg=-mfpu=neon",
+  "-C", "link-arg=-mfloat-abi=hard",
+  "-C", "link-arg=-mcpu=cortex-a9",
+  "-C", "link-arg=--sysroot=<toolchain_install_path>/sysroots/cortexa7hf-neon-remarkable-linux-gnueabi",
+]
+```
+
 You can also add this snippet to the above file in order to default to cross-compiling for this project:
 
 ```
--- a/README.md
+++ b/README.md
@@ -23,23 +23,9 @@ In order to build `libremarkable` and the examples (`spy.so` and `demo`), you'll
 
 You can then set up your Rust toolchain for cross compilation with: `rustup target add armv7-unknown-linux-gnueabihf`.
 
-Once that's done, you should add the following to `.cargo/config` (replace `<path-to-installed-oecore-toochain>` with the directory you installed the Remarkable toolchain to):
-```
-[target.armv7-unknown-linux-gnueabihf]
-linker = "<path-to-the-installed-oecore-toolchain>/sysroots/x86_64-oesdk-linux/usr/bin/arm-oe-linux-gnueabi/arm-oe-linux-gnueabi-gcc"
-rustflags = [
-  "-C", "link-arg=-march=armv7-a",
-  "-C", "link-arg=-marm",
-  "-C", "link-arg=-mfpu=neon",
-  "-C", "link-arg=-mfloat-abi=hard",
-  "-C", "link-arg=-mcpu=cortex-a9",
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,23 +23,9 @@ In order to build `libremarkable` and the examples (`spy.so` and `demo`), you'll
...
-  "-C", "link-arg=--sysroot=<path-to-the-installed-oecore-toolchain>/sysroots/cortexa9hf-neon-oe-linux-gnueabi",
-]
-```
-
-(`<path-to-the-installed-oecore-toolchain` will likely be `/usr/local/oecore-x86_64/`, if you did the default install on Linux.)
-
-If you have further questions, feel free to ask in Issues.
+In order for rust to leverage the toolchain a `.cargo/config` file is required. This file can be generated using the `gen_cargo_config.py`. First the toolchain environment must be
+sourced. It location is can be found within the toolchain installation directory. The correct path is also referenced in the toolchain [wiki](https://remarkablewiki.com/devel/toolchain).
+After the environment is loaded the script will read the environment variables to generate the correct `.cargo/config` file for your toolchain.
```
bkirwi, line 28: I suggest adding an example here to, like:

> For example, when run against toolchain version 1.2.3, this might generate:
>
> ```
> <config file text>
> ```

Aside from documenting the format, this may be useful to folks who don't want to or can't run the script.

gen_cargo_config.py
```diff
@@ -0,0 +1,38 @@
...
+    extract_compiler_path = re.compile(re_parse_path)
+    toolchain_compiler_path = extract_compiler_path.search(os.environ["PATH"])[0][1:-1]
+    #join path to get full compiler path
+    cc = os.path.join(toolchain_compiler_path, cc)
+    
+    #create .cargo dir if not exsits
+    if not os.path.isdir(".cargo"):
+        os.mkdir(".cargo")
+    with open(".cargo/config","w") as f:
+        f.write(cargo_config.format(gcc_path=cc, sysroot=sysroot))
```
bkirwi, line 35: It would be kind to print a success message here.

bkirwi (commented):
A few suggestions, but overall great!

gen_cargo_config.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python
```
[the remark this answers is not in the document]
LinusCDE: On that note, a python version should probably be specified. I wasn't sure if ubuntu still defaulted to python so i tested it out by installing python 2 and 3 into a ubuntu:20.04 docker container. At least in the container, it doesn't default to any version, making the "python" command non-existant.

I personally also prefer using `#!/usr/bin/env python3` (if 3 is chosen) since a python installation from source would for example be installed to `/usr/local/bin/` instead. This makes `env` find the first working python3 binary according to the path and I never stumbled upon any installation that didn't have `env` installed/symlinked in that location.

LinusCDE (commented):
The script seems fine so far. Just added a few little things below this comment.

I also tested out whether all toolchains work, and it only worked for oecore for me:

grafik

I used the source commands I added to the file for each respective toolchain. They all run into your exact issue #101 and attempted to build the example in dev mode after running the script `gen_cargo_config.py`.

The `.cargo/config` for poky was the exact same, I had [used earlier in my projects](https://github.com/LinusCDE/retris/blob/master/.cargo/config). So I would suspect that this is due to some legacy issue??

For the codex toolchains I now confused myself as well. I ensured that the paths all exist (installed to standard locations), so it's probably not a typo.

grafik

Any suggestions on this?

Edit: A few more informations on my test-setup:
 - Command used to build: cargo build --target=armv7-unknown-linux-gnueabihf --example demo (respective env file was source before)
 - Using a recent Manjaro Linux. All 4 toolchains are installed to their standard locations.

ju6ge:
Hm so I tested the script for the oecore (latest) and codex (3.1.15) version installed on my system. Both work fine in release as well as in dev mode. :thinking: Has anyone else tried using the script yet?

I can test this in more detail on the weekend. Currently I need to focus on getting my master thesis submitted this week :see_no_evil: :laughing:

LinusCDE:
> I can test this in more detail on the weekend. Currently I need to focus on getting my master thesis submitted this week

Nice! Don't put yourself up with this then. We're not in a hurry!

ju6ge:
Okay so I submitted my thesis and now I have some time for a closer look. I took my laptop as a clean environment where I have not built libremarkable previously. I have installed all 4 toolchains: poky, oecore, codex: rm10x and rm11x. For every toolchain I am sourcing the environment file and running the generation script, afterwards I build the demo binary in dev mode:

Codex rm11x: Builds ✅
2022-04-15T14:20:05,227852329+02:00

Codex rm10x: Builds ✅
2022-04-15T14:21:51,542741975+02:00

Oecore: Builds ✅
2022-04-15T14:46:00,916547729+02:00

Poky: Builds ✅
2022-04-15T15:05:21,185605384+02:00

I could not find a problem with any of the toolchains in a clean environment. So I am a bit confused why they don't seem to work in your setup. Any ideas?

LinusCDE:
> I could not find a problem with any of the toolchains in a clean environment. So I am a bit confused why they don't seem to work in your setup. Any ideas?

Not sure myself tbh. I checked against typos at least and poky definitely used to work as I used it for a really long time. Maybe it's some rust toolchain changes and or lib incompatibilities in my system.

Since you've shown that all toolchains work, I don't have anything against merging this now. Using this, is a far better system than before. I'll probably do the same in my repos, and drop a quick explanation + link to this python script in the readme of those projects. I'm up to merging this as-is.

Anything to add/change in your opinion @bkirwi ? Otherwise I'll just merge this in 2-3 days or so.

LinusCDE (approved):

fenollp (approved):

fenollp:
Off topic: what kind of advantage does one get by using one of the possible toolchains instead of a somewhat generic one with `cross`? 

If it's about linking with the right libs/versions, does musl present more disadvantages than pros in your experience?

Thanks for your work

LinusCDE:
Not sure if this was a general question or one at @ju6ge .

Imo, I don't see a real benefit for most rust applications as well.

If I had to guess it would be stuff like:
 - Performance benefits and possibly more optimizations with a more recent gcc (e.g. for system libs like zstd-sys)
 - If some rust dependency would link against a shared library, "generic" toolchains only have the bare minimum or possibly wrong version that are not present on the system

The second point usually only seems to highly matter for qt libs. You could use it in rust, but I have not seen anyone using it yet. And openssl is also only shipped as a binary which necessitates vendoring the lib anyway.

GCC version of all the toolchains:

grafik

Please correct me if anything as missed or misrepresented. (I usually prefer an older toolchain as this prevents glibc version errors which I had plenty for work related stuff lately. That forces a binary only to run on very recent distro versions.)

bkirwi:
No objections here! Since we're keeping the example in the README this should have no downsides.

> Off topic: what kind of advantage does one get by using one of the possible toolchains instead of a somewhat generic one with cross?

Personal example: some of my code depends on the `tflite` crate, and cross' built-in toolchains can't build that out of the box. I think any env vars etc. you may need to set are a bit more awkward with cross as well. I suspect these are solvable problems... but so far I've had an easier time using reMarkable's toolchain, and installing the other stuff I needed manually, than building a custom docker container with everything I've needed.

ju6ge:
I really only used `cross` once, so not sure … I think the benefit lies more in the fact that you can use it in situations where you do not have access to an official toolchain. Apart from maybe a streamlined installation path and build environment. 🤷‍

PR #103, opened 2022-04-17
Fix submodule URL for Freetype

matteodelabre:
Due to recent [changes in GitHub’s supported protocols](https://github.blog/2021-09-01-improving-git-protocol-security-github/), the fetching of submodules of this repository is failing. This also causes [folly](https://github.com/bkirwi/folly)’s build to fail, since for some reason the submodules are fetched as part of the build (@bkirwi).

This patch simply updates the submodule’s URL to use https:// instead of git://.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "legacy-c-impl/libfreetype"]
 	path = legacy-c-impl/libfreetype
-	url = git://github.com/ricardoquesada/libfreetype
+	url = https://github.com/ricardoquesada/libfreetype


bkirwi:
> for some reason the submodules are fetched as part of the build

This is `cargo`'s default behaviour. Normally helpful, but annoying for this repo, since the freetype dep is only for the deprecated C version of libremarkable.

(@fenollp / @LinusCDE - how would you feel about removing the C code from the main branch? If you're open to the idea I can probably do the work of splitting it out.)

@matteodelabre - assuming this works I'll make sure to percolate it through to the dependents, and make the toltec PR with the new rev.

LinusCDE:
> ([...] - how would you feel about removing the C code from the main branch? If you're open to the idea I can probably do the work of splitting it out.)

I have no real opinion on this. I don't know the history one it, or who might still use it.

From my point of view, you can split it out, but we should at least point out where it went in the readme, since people might search it here and wonder where it went.

fenollp:
How about just renaming `.gitmodules`

PR #104, opened 2022-04-19
Remove the legacy C implementation from the source tree

bkirwi:
This branch:
- Removes the `legacy-c-impl` subdirectory and the libfreetype submodule.
- Updates the readme to point to a branch where the `legacy-c-impl` is still present.

@LinusCDE - IIUC this is more or less the approach you were suggesting.

@fenollp - definitely better than the status quo, but it would prevent the legacy c impl even from building without changes, and this seems like a nice chance to declutter the repo. I don't feel that strongly though!


LinusCDE (approved):
Looks good to me 👍

fenollp (approved):

bkirwi:
Thanks all!

Issue #106, opened 2022-09-01
evdev type mismatch


glyh:
I've got this error message for evdev, when I was running `make TARGET=armv7-unknown-linux-gnueabihf deploy-x-demo`:
```
error[E0308]: mismatched types
   --> /cargo/registry/src/github.com-1ecc6299db9ec823/evdev-0.11.7/src/lib.rs:390:29
    |
390 |     let dur = Duration::new(tv.tv_sec.unsigned_abs(), tv.tv_usec as u32 * 1000);
    |               ------------- ^^^^^^^^^^^^^^^^^^^^^^^^ expected `u64`, found `u32`
    |               |
    |               arguments to this function are incorrect
    |
note: associated function defined here
   --> /rust/lib/rustlib/src/rust/library/core/src/time.rs:185:18
    |
185 |     pub const fn new(secs: u64, nanos: u32) -> Duration {
    |                  ^^^
help: you can convert a `u32` to a `u64`
    |
390 |     let dur = Duration::new(tv.tv_sec.unsigned_abs().into(), tv.tv_usec as u32 * 1000);
    |                                                     +++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `evdev` due to previous error
```

LinusCDE:
Hi.

At first I couldn't reproduce this error. But then noticed that your version of evdev was newer then the one in my (older) `Cargo.lock` file.

Seems something between evdev 0.11.3 and 0.11.7 changed that broke it. I'm looking into it and will create a PR to fix it shortly.

Good catch! :+1:

LinusCDE:
The same compilation error was found 9 days ago in issue https://github.com/emberian/evdev/issues/95 (which lead to the PR fixing it)

LinusCDE:
The issue at hand is fixed in the [0.6.1 release](https://crates.io/crates/libremarkable/0.6.1).

As said, when the new evdev release is around, we can update to it and close this issue. That will then be included in a future `0.6.2` or  `0.7.0` release when we get to do that (not making that release for this fix alone though).

PR #107, opened 2022-09-01
Prevent usage newer evdev than 0.11.3 due to compilation error for armv7

LinusCDE:
This should fix issue #106 by capping the version of evdev used for now.

I also tested with the newer minor version (0.12.0) and it had the same issue as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ image = { version = "0.23.14", optional = true }
 line_drawing = { version = "1.0.0", optional = true }
 
 # input
-evdev = { version = "0.11.3", optional = true }
+evdev = { version = "<=0.11.3", optional = true }
 epoll = { version = "4.3.1", optional = true }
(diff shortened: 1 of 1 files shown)


LinusCDE:
Was about to make a PR to evdev to fix the source issue, but found that their current master branch builds fine. It seems this already got addressed in a commit 9 days ago, but there is no release with this fix yet.

PR: https://github.com/emberian/evdev/pull/96
Commit: https://github.com/emberian/evdev/commit/909b3d2f53fae43f5ed42ece5a22d37a797edf38

So this fix should do the trick until the next release of evdev fixes the issue. If it's a patch release (`0.12.X`) we should probably use `>=` for the version to ensure no accidental use of `0.12.0`.

Will merge this tomorrow of no one is against this (cc @bkirwi @fenollp )

bkirwi:
I'm all for it - thanks for handling!

fenollp (approved):

PR #114, opened 2023-07-23
Update wiki links to archive.org

Eeems:
The wiki is no longer loading, but luckily there are recent archive.org crawls of the site that can be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,12 @@ For further documentation see the [wiki](https://github.com/canselcik/libremarka
 ### Build Instructions
 
 #### Setting up the toolchain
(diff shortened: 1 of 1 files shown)


LinusCDE:
The wiki being discontinued is pretty sad. Do you know by any chance who hosts it? Maybe we could continue hosting it somewhere. Would have no issues doing that myself if needed.

Eeems:
@LinusCDE There is an email address on https://web.archive.org/web/20230331220740/https://remarkablewiki.com/ for someone claiming to be the owner. That said, I think it's probably a better idea to have people transpose the information to remarkable.guide

PR #116, opened 2024-01-02
reduce required dependencies

HookedBehemoth:
- move env_logger to dev-dependencies (was only used in the examples)
- move image functions behind existing feature flag

I bumped the version to 0.7.0 as this could be a breaking change for someone having appctx but not image enabled.

The motivation was to flatten the dependency tree, reduce build time and reduce target folder size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libremarkable"
-version = "0.6.2"
+version = "0.7.0"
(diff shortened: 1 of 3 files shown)


bkirwi:
Seems reasonable enough to me! I'm not aware of any real-world applications that use `appctx`, but it still seems wise to bump the version.

@fenollp / @LinusCDE - unless you object this seems mergeable?

fenollp:
LGTM!

Issue #117, opened 2024-01-19
doesn't work with timower's rm2fb


HookedBehemoth:
Presumably because it switched from message queues to unix sockets and libremarkable has it's own IPC code instead of using rm2fb-client.so.
https://github.com/timower/rM2-stuff/tree/master/libs/rm2fb
> Uses UNIX sockets instead of message queues. Makes it easier to implement synchronized updates.
```
thread 'main' panicked at 'Failed to open swtfb shared buffer: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libremarkable-0.6.2/src/framebuffer/core.rs:128:18
```

I'll look into it further.

Eeems:
It sounds like timower's rm2fb broke backwards compatibility with apps and should re-add message queue support for older clients.

LinusCDE:
I'm not aware of his library, but yeah our default behaviour for the Gen 2 model is to use our own client. It should offer all the same functionality as the official rM2FB shim while making it more clear to developers that they're using one and giving them more control.

However that means that updates to rM2FB could break existing binaries with libremarkable and had been noted by rM2FB developers before.

If timowers rm2fb is acting like a shim and using different means to communicate with its server, that would certainly confuse libremarkable as it would behave like the official rM2FB shim, but not find the correct server via it's usual means.

In this case, you can as the app developer, choose to use the old method even on the rM 2. For that use the construct [`Framebuffer::device`](https://docs.rs/libremarkable/latest/libremarkable/framebuffer/core/struct.Framebuffer.html#method.device) instead of [`Framebuffer::new`](https://docs.rs/libremarkable/latest/libremarkable/framebuffer/core/struct.Framebuffer.html#method.new) and it should work better on the rM 2 then and resolve your issue. You can get the parameter for `device()` from [`device::CURRENT_DEVICE::get_framebuffer_path`](https://docs.rs/libremarkable/latest/libremarkable/device/struct.Device.html#method.get_framebuffer_path) or just look at the source code for the `new()` function mentioned above.

timower:
re-implementing the rm2fb protocol in libremarkable seems very fragile. As soon as the protocol changes apps will start to fail.

Eeems:
> re-implementing the rm2fb protocol in libremarkable seems very fragile. As soon as the protocol changes apps will start to fail.

[...]

PR #118, opened 2024-01-26
Allow disabling the automatic use of our internal swtfb client by setting an env

LinusCDE:
There was an issue raised at #117 where presumably a new library is used as a shim for rM 2 Framebuffer support, but it doesn't seem to use the offical server, which our client relies on.

In general there had been issues raised that issues could arise even with the official shim if it gets updated but our implementation still does the old behaviour. Such applications would require to be recompiled to use `Framebuffer::device` instead of `Framebuffer::new` or otherwise not work.

This PR allows people to make libremarkable choose the old implementation in case the developer didn't care about it anyway / left the choice up to libremarkable.

It is done by setting the lengthy env `LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB` to `1`, `yes` or `true` (case-insensitive). Feel free to suggest a better name, I couldn't think of any.

Package maintainers could also fix arising issues by setting that env in some wrapper script for that binary without the need of any source code change or even re-complication if the binary is recent enough.

Basically this should future proof applications built using libremarkable better.

### DIFF
--- a/src/framebuffer/core.rs
+++ b/src/framebuffer/core.rs
@@ -50,7 +50,16 @@ impl Framebuffer {
         let device = &*device::CURRENT_DEVICE;
         match device.model {
             Model::Gen1 => Framebuffer::device(device.get_framebuffer_path()),
-            Model::Gen2 => Framebuffer::rm2fb(device.get_framebuffer_path()),
+            Model::Gen2 => {
+                // Auto-select old method still if env LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB is set affirmatively
+                if let Ok(env_answer) = std::env::var("LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB") {
+                    let env_answer = env_answer.to_lowercase();
+                    if env_answer == "1" || env_answer == "true" || env_answer == "yes" {
+                        Framebuffer::device(device.get_framebuffer_path())
+                    }
+                }
+                Framebuffer::rm2fb(device.get_framebuffer_path())
+            }
         }
     }
 


src/framebuffer/core.rs
```diff
@@ -50,7 +50,19 @@ impl Framebuffer {
         let device = &*device::CURRENT_DEVICE;
         match device.model {
             Model::Gen1 => Framebuffer::device(device.get_framebuffer_path()),
-            Model::Gen2 => Framebuffer::rm2fb(device.get_framebuffer_path()),
+            Model::Gen2 => {
+                // Auto-select old method still if env LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB is set affirmatively
```
fenollp, line 54: Could you add a mention in the README about this behavior plz?

src/framebuffer/core.rs
```diff
@@ -50,7 +50,19 @@ impl Framebuffer {
...
+                if let Ok(env_answer) = std::env::var("LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB") {
+                    let env_answer = env_answer.to_lowercase();
+                    if env_answer == "1" || env_answer == "true" || env_answer == "yes" {
+                        Framebuffer::device(device.get_framebuffer_path())
+                    } else {
+                        Framebuffer::rm2fb(device.get_framebuffer_path())
+                    }
+                } else {
+                    Framebuffer::rm2fb(device.get_framebuffer_path())
+                }
```
fenollp, line 64: How about this? (untested)
```suggestion
                match std::env::var("LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB").ok() {
                    Some("1") => Framebuffer::device(device.get_framebuffer_path()),
                    _ => Framebuffer::rm2fb(device.get_framebuffer_path()),
                }
```

fenollp: (I'm more inclined to be restrictive in the possible values, but feel free to revert that)

Havner:
Hi,

I can't get this to work as intended. Seems that libremarkable doesn't like rm2fb client. I've compile the demo and it works properly (through built-in client). But with this it doesn't work. Has anybody managed to make it work?

(Software 2.15, original rm2fb from toltec, not timower)

```
reMarkable: /opt/demo/ LD_PRELOAD=/opt/lib/librm2fb_client.so.1 LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB=1 ./demo
thread 'main' panicked at src/framebuffer/core.rs:206:9:
FBIOGET_VSCREENINFO failed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
reMarkable: /opt/demo/ LIBREMARKABLE_FB_DISFAVOR_INTERNAL_RM2FB=1 /opt/bin/rm2fb-client ./demo
thread 'main' panicked at src/framebuffer/core.rs:206:9:
FBIOGET_VSCREENINFO failed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #119, opened 2024-01-26
Fixed Crashing Due To Unknown Pen Inputs

Ap6661:
Unanticipated inputs from the Wacom Pen will cause the code to reach unreachable causing the entire app to crash. This includes any side buttons on 3rd party pens

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -548,7 +548,7 @@ fn on_wacom_input(app: &mut appctx::ApplicationContext<'_>, input: input::WacomE
                         wacom_stack.clear();
                     }
                 }
-                _ => {},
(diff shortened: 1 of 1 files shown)


examples/demo.rs
```diff
@@ -548,7 +548,7 @@ fn on_wacom_input(app: &mut appctx::ApplicationContext<'_>, input: input::WacomE
                         wacom_stack.clear();
                     }
                 }
-                _ => unreachable!(),
+                _ => {},
```
fenollp, line 551:
```suggestion
                _ => {}
```

fenollp (changes requested):

Issue #122, opened 2024-04-10
More intuitive color use for the draw_text function


FeDorDrolma98:
Hello, 
The draw_text function currently use this formula to modify the color : 
`let mult = (1.0 - v).min(1.0);
color::RGB((c1 * mult) as u8, (c2 * mult) as u8, (c3 * mult) as u8)`

But when using it, it modify only the background of the glyph (which stay in black) :
<img src="https://github.com/canselcik/libremarkable/assets/58807436/feb347f3-c82e-48ed-b6b5-2bf45c63fd15" width=25% height=25%>

Since it is more interesting to modify the color of the glyph itself, maybe this formula could be better (that's what I'm using) :
`color::RGB((255.0+(c1-255.0)*v ) as u8, (255.0+(c2-255.0)*v  ) as u8, (255.0+(c3-255.0)*v  ) as u8)`
If we want it to work we also have to modify this (not useful anymore to do '255 - components[_]' I guess):
```
        let c1 = f32::from(components[0]);
        let c2 = f32::from(components[1]);
        let c3 = f32::from(components[2]);
```

Then it is really the font that is changing color :
<img src="https://github.com/canselcik/libremarkable/assets/58807436/73469246-1caa-4b42-9399-af05f904273f" width=25% height=25%>

What do you think?

andrewjensen:
I would love having independent `color_foreground` and `color_background` arguments. That would allow for more faded text like the screenshot above as well as text on darker panels.

PR #123, opened 2024-06-19
More intuitive color use for the draw_text function

gamirab:
potentially, add background (this is for foreground color)
See issue #122


fenollp:
Sounds good to me. Please clean up this patch though: remove dead code and format with `cargo fmt`.

I'll send a PR with fixes for the unrelated Clippy warnings.

gamirab:
Sorry for that ! And thanks !

fenollp (approved):

PR #125, opened 2024-07-27
Bump deps

fenollp:
I bumped all the deps except `libc` and:
* [`once_cell`](https://github.com/canselcik/libremarkable/pull/129) can be dropped, if we're alright requiring rust >=1.80
* [`aabb-quadtree`](https://github.com/canselcik/libremarkable/pull/126) is causing me compilation issues...
* [`atomic`](https://github.com/canselcik/libremarkable/pull/127) requires some work (I'd say `impl From<std::sync::atomic::AtomicI64> for DrawMode`, allowing to drop `atomic`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,26 +10,26 @@ exclude = [ "reference-material/*", "legacy-c-impl/*", "private/*" ]
 edition = "2021"
 
 [dependencies]
-log = "0.4.14"
-once_cell = "1.9.0"
+log = "0.4.22"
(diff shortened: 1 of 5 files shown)


bkirwi (commented):
Looking it over, it seems like stopwatch is only used in a very limited way that would be quite straightforward to replace with the stdlib `Instant::elapsed` etc. -- might be nicer than depending on a git rev here? (I can take a stab at that change if it's a hassle...)

PR #127, opened 2024-07-27
Drop `atomic` crate

fenollp:


bkirwi (approved):
All looks reasonable - thanks!

PR #129, opened 2024-07-27
deps: drop once_cell with MSRV=1.80

fenollp:


LinusCDE:
Finally fixed. Will do a quick sanity check on 3 of my apps, update the readme with the MSRV requirement, then release 0.7.0.

PR #131, opened 2024-09-09
Fix a couple recent regressions

bkirwi:
First: when we added an env var recently to allow forcing a fallback to the rm1 framebuffer interface, we were still using the rm2 path. This path does exist, but attempting ioctls on it will fail. Fixed by specifying the old path when setting the env var.

Second: this recent font rendering change broke the demo app! I didn't look into it, but for now let's just get main back in a good state. It's a small diff.


bkirwi:
I did a small about of manual testing with these two changes; setting the env var seems to do the right thing now and the demo app renders text correctly.

fenollp (approved):
LGTM!

bkirwi:
Thanks!

